PR #57, opened 2026-01-03
Support merge RobotCfg 

yuecideng:
# Description

- Support merge `RobotCfg` from a given dict
- Modify `DexforceW1Cfg` to support merge user define urdf components
- Fix the issue of regex pattern of default hand in `DexforceW1Cfg`

## Type of change

## Example

Check file in `examples/sim/robot/dexforce_w1.py`

## Checklist

### DIFF
--- a/embodichain/lab/sim/cfg.py
+++ b/embodichain/lab/sim/cfg.py
@@ -1036,9 +1036,9 @@ class RobotCfg(ArticulationCfg):
             keys corresponding to the control parts name.
         - The joint names in the control parts support regular expressions, e.g., 'joint[1-6]'.
             After initialization of robot, the names will be expanded to a list of full joint names.
-        - `Robot` is a derived class of `Articulation`, with contorl parts support. So the `drive_pros`
+        - `Robot` is a derived class of `Articulation`, with control parts support. So the `drive_pros`
             in `ArticulationCfg` can use control part as key to specify the corresponding joint drive properties, 
-            which will be orverridden if these joint names are already specified.
+            which will be overridden if these joint names are already specified.
     """
 
     urdf_cfg: URDFCfg | None = None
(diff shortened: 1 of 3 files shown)


embodichain/lab/sim/robots/dexforce_w1/cfg.py
```diff
@@ -50,15 +50,16 @@ class DexforceW1Cfg(RobotCfg):
 
     version: DexforceW1Version = DexforceW1Version.V021
     arm_kind: DexforceW1ArmKind = DexforceW1ArmKind.INDUSTRIAL
+    with_default_eef: bool = True
 
     @classmethod
     def from_dict(
-        cls, init_dict: typing.Dict[str, typing.Union[str, float, tuple]]
+        cls, init_dict: Dict[str, str | float | tuple | dict]
```
chase6305, line 57: Currently, does embodichain no longer support Python versions 3.8 or below 3.9? If you use `Dict[str, List[str]] | None` in Python 3.8 or 3.9, you will get a syntax error because this type union syntax should only be supported from Python 3.10 onwards.

yuecideng: Yes, we will drop support for older python version below py310. The reasons including:
- compatible with numpy 2.0
- compatible with lerobot 
- compatible with latest warp-lang

examples/sim/robot/dexforce_w1.py
```diff
@@ -0,0 +1,75 @@
...
+            },
+            "urdf_cfg": {
+                "components": [
+                    {
+                        "component_type": "left_hand",
+                        "urdf_path": "DH_PGC_140_50/DH_PGC_140_50.urdf",
+                    },
+                    {
+                        "component_type": "right_hand",
+                        "urdf_path": "DH_PGC_140_50/DH_PGC_140_50.urdf",
```
chase6305, line 52: You could consider using the `DH_PGC_140_50_M` assets. Dexforce specifically modified the finger and connector models, and the joints are now standardized as `LEFT_FINGER1`, `LEFT_FINGER2`, `RIGHT_FINGER1`, and `RIGHT_FINGER2`.

chase6305 (approved):

PR #63, opened 2026-01-13
fix refactor problems

yhnsu:
# Description
fix refactor problems
- [x] add grasp_pose_object attr to RigidObjectCfg to avoid bug
- [x] fix ScoopIce Env bug


yuecideng (approved):

embodichain/lab/gym/envs/managers/datasets.py
```diff
@@ -295,11 +285,6 @@ def _initialize_dataset(self) -> None:
         fps = self.robot_meta.get("control_freq", 30)
         features = self._build_features()
 
-        logger.log_info("------------------------------------------")
```
yuecideng, line 298: Whywas the logging info removed?

embodichain/lab/gym/envs/managers/datasets.py
```diff
@@ -36,16 +36,9 @@
 if TYPE_CHECKING:
     from embodichain.lab.gym.envs import EmbodiedEnv
 
-try:
```
yuecideng, line 39: Why was the import guard removed?

embodichain/utils/visualizer.py
```diff
@@ -148,6 +148,241 @@ def draw_feature(
     return vis_features
 
 
+class HeatMapEnvV3:
```
yuecideng: This class can be implemented internally for now

embodichain/lab/gym/envs/managers/datasets.py
yhnsu: > Why�was the logging info removed?

because it's redundant. The same informations are already logged elsewhere.

PR #80, opened 2026-01-20
RL env with action preprocess

yhnsu:
# Description

- [x] RL env for easier implementation for new tasks
- [x] action preprocessor supported different type of action input
- [x] update RL docs

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,13 +48,12 @@ jobs:
       - uses: actions/checkout@v4
       - name: Build docs
         run: |
-          pip install -e .[lerobot] --extra-index-url http://pyp.open3dv.site:2345/simple/ --trusted-host pyp.open3dv.site
+          pip install -e . --extra-index-url http://pyp.open3dv.site:2345/simple/ --trusted-host pyp.open3dv.site
           pip install -r docs/requirements.txt
           cd ${GITHUB_WORKSPACE}/docs
           echo "Start Building docs..."
           pip uninstall pymeshlab -y
           pip install pymeshlab==2023.12.post3
-          pip install matplotlib==3.10.8
           make html
       - name: Upload docs artifact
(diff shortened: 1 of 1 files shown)


embodichain/lab/gym/envs/managers/rewards.py
```diff
@@ -30,7 +30,7 @@
 def distance_between_objects(
     env: EmbodiedEnv,
     obs: dict,
-    action: torch.Tensor,
+    action: dict,
```
yuecideng: `torch.Tensor | dict` would be better?

embodichain/lab/gym/envs/managers/rewards.py
```diff
@@ -171,28 +171,34 @@ def action_smoothness_penalty(
...
+            break
+
     # Use dictionary-based state management
     if not hasattr(env, "_reward_states"):
         env._reward_states = {}
 
     # compute difference between current and previous action
     if "prev_actions" in env._reward_states:
-        action_diff = action - env._reward_states["prev_actions"]
+        action_diff = action_tensor - env._reward_states["prev_actions"]
```
yuecideng: The `_reward_states` maybe replaced by historical obs/action (like `episode_obs_list` or a customized length buffer)

embodichain/lab/gym/envs/rl_env.py
```diff
@@ -0,0 +1,301 @@
...
+
+        Args:
+            action: End-effector pose (position + orientation)
+                   Shape: (num_envs, 6) for pos+euler or (num_envs, 7) for pos+quat
+
+        Returns:
+            Joint positions from IK
+        """
+        # Auto-select control part if not specified
+        control_part_name = self.control_part_name
```
yuecideng: `control_part` is an optional attribute of a robot. We should also support this case for eef pose conversion

PR #88, opened 2026-01-23
Articulation support getting target qpos

matafela:
# Description

* `Articulation` now support getting target qpos
* Fixed a bug in the CUDA environment where setting some joint angles would affect other joint angles.

## Type of change

## Checklist

### DIFF
--- a/docs/source/overview/sim/sim_articulation.md
+++ b/docs/source/overview/sim/sim_articulation.md
@@ -14,7 +14,12 @@ Articulations are configured using the {class}`~cfg.ArticulationCfg` dataclass.
 | `init_pos` | `tuple` | `(0,0,0)` | Initial root position `(x, y, z)`. |
 | `init_rot` | `tuple` | `(0,0,0)` | Initial root rotation `(r, p, y)` in degrees. |
 | `fix_base` | `bool` | `True` | Whether to fix the base of the articulation. |
+| `init_qpos` | `List[float]` | `None` | Initial joint positions. |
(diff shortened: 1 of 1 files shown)


embodichain/lab/sim/objects/articulation.py
```diff
@@ -248,6 +251,34 @@ def qpos(self) -> torch.Tensor:
...
+        else:
+            self.ps.gpu_fetch_joint_data(
+                data=self._target_qpos,
+                gpu_indices=self.gpu_indices,
+                data_type=ArticulationGPUAPIReadType.JOINT_TARGET_POSITION,
+            )
+            return self._target_qpos[:, : self.dof].clone()
+
     @property
     def qvel(self) -> torch.Tensor:
```
yuecideng, line 285: Maybe also add `target_qvel`

embodichain/lab/sim/objects/articulation.py
```diff
@@ -979,7 +1010,7 @@ def set_qpos(
                 self.body_data.qpos
```
yuecideng: We should replace `self.body_data.qpos` to `self.body_data.target_qpos` here

embodichain/lab/sim/objects/articulation.py
```diff
@@ -976,10 +1007,10 @@ def set_qpos(
         else:
             # TODO: trigger qpos getter to sync data, otherwise crash
             if joint_ids is not None:
-                self.body_data.qpos
+                self.body_data.target_qpos
```
yuecideng: Also need to check whether `target=True` is applied. If `target` is False, we can apply `qpos` to articulation

PR #119, opened 2026-02-04
save success status before reset

yhnsu:
# Description

buf fix: save success status before reseting objects status

### DIFF
--- a/embodichain/lab/gym/envs/embodied_env.py
+++ b/embodichain/lab/gym/envs/embodied_env.py
@@ -587,20 +587,6 @@ class EmbodiedEnv(BaseEnv):
             "The method 'create_demo_action_list' must be implemented in subclasses."
         )
 
-    def is_task_success(self, **kwargs) -> torch.Tensor:
-        """
-        Determine if the task is successfully completed. This is mainly used in the data generation process
-        of the imitation learning.
-
-        Args:
-            **kwargs: Additional arguments for task-specific success criteria.
-
-        Returns:
-            torch.Tensor: A boolean tensor indicating success for each environment in the batch.
-        """
-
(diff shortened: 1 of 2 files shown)


embodichain/lab/gym/envs/embodied_env.py
```diff
@@ -379,7 +379,13 @@ def _initialize_episode(
         if save_data and self.cfg.dataset:
             if "save" in self.dataset_manager.available_modes:
 
-                current_task_success = self.is_task_success()
+                # Use task success status saved before reset_objects_state
+                current_task_success = getattr(self, "_task_success", None)
```
yuecideng: We now always has `_task_success` attribute, so we can use it directly

yuecideng (changes requested):

embodichain/lab/scripts/preview_env.py
```diff
@@ -86,7 +86,18 @@
     ##############################################################################################
     # load gym config
     gym_config = load_json(args.gym_config)
-    cfg: EmbodiedEnvCfg = config_to_cfg(gym_config)
+
+    # Define manager modules for config parsing
+    manager_modules = [
```
yuecideng: It would be better to define a functor to return the default manager modules. Then you can just define once and reused for different parts

yuecideng (changes requested):

pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
   "pin-pink",
   "casadi",
   "qpsolvers==4.8.1",
+  "qpsolvers[osqp]",
```
yuecideng: why not combine these two lines?

yuecideng (approved):

yuecideng (approved):

PR #142, opened 2026-02-18
Split robot left right index in semantic mask computation

dexscutai:
image

### DIFF
--- a/embodichain/lab/gym/envs/managers/observations.py
+++ b/embodichain/lab/gym/envs/managers/observations.py
@@ -38,6 +38,7 @@ def get_rigid_object_pose(
     env: EmbodiedEnv,
     obs: EnvObs,
     entity_cfg: SceneEntityCfg,
+    to_matrix: bool = True,
 ) -> torch.Tensor:
     """Get the world poses of the rigid objects in the environment.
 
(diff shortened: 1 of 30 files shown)


embodichain/lab/gym/envs/managers/observations.py
```diff
@@ -180,10 +180,11 @@ def compute_semantic_mask(
     """Compute the semantic mask for the specified scene entity.
 
     Note:
-        The semantic mask is defined as (B, H, W, 3) where the three channels represents:
-        - robot channel: the instance id of the robot is set to 1 (0 if not robot)
+        The semantic mask is defined as (B, H, W, len(SemanticMask)) where these channels represents:
```
yuecideng, line 219: Consider use a more general and configurable method todefine the mapping from object | links uid to semantic index.

For example, use a dict as below:
```
{
    "background": 0,
    "foreground": {
          "obj1": 1,
          "onj2": 2,
    },
    "robot": {
          "right_*": 3,
          "left_*": 4
     }
}
```

yuecideng (changes requested):

embodichain/lab/gym/envs/managers/observations.py
dexscutai: UID has been defined in enum file. Now it is sufficient to use for algorithm recently.

PR #152, opened 2026-02-28
fix opw solver

matafela:
# Description

- Fix opw cpu batch inverse kinematic qpos seed problem. (use warp cpu)
- Fix opw gpu inverse kinematic.
- Add more test case in opw unittest.

## Type of change

## Checklist

### DIFF
--- a/embodichain/utils/warp/kinematics/opw_solver.py
+++ b/embodichain/utils/warp/kinematics/opw_solver.py
@@ -436,8 +436,6 @@ def opw_ik_kernel(
             qpos[qpos_start + k] = normalize_to_pi(
                 (theta[idx] + offsets[k]) * sign_corrections[k]
             )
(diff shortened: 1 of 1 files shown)


embodichain/lab/sim/solvers/opw_solver.py
```diff
@@ -121,10 +121,9 @@ def __init__(self, cfg: OPWSolverCfg, device: str = "cpu", **kwargs):
 
         """
         super().__init__(cfg=cfg, device=device, **kwargs)
-        if self.device.type == "cpu":
-            self._init_py_opw_kinematics_solver(cfg, **kwargs)
-        else:
-            self._init_warp_solver(cfg, **kwargs)
+        # if self.device.type == "cpu":
```
yuecideng: We may just remove this part of code.

PR #229, opened 2026-04-13
Fix opw solver

matafela:
# Description

- Fix opw ik solver in cpu mode.
- Fix gpu memory out of bound if directly using opw solver.
- Add safe_margin to OPW solver cfg.
- SolverCfg supports qpos limits now.

TODO:
- [x] Findout what operator in warp cpu failed.

## Type of change

## Checklist


yuecideng (approved):

embodichain/lab/sim/solvers/base_solver.py
```diff
@@ -72,6 +72,9 @@ class SolverCfg:
     when multiple solutions are available.
     """
 
+    qpos_limits: List[float] | None = None
```
yuecideng: Use `user_qpos_limits` would be better. Also add docs to indicate the behaviour of None.

embodichain/lab/sim/objects/robot.py
```diff
@@ -934,6 +934,9 @@ def init_solver(self, cfg: Union[SolverCfg, Dict[str, SolverCfg]]) -> None:
                     ):
                         solver_cfg.joint_names = self.cfg.control_parts[part_name]
                     self._solvers[name] = solver_cfg.init_solver(device=self.device)
+                    joint_ids = self.get_joint_ids(name=part_name)
+                    joint_limits = self._data.qpos_limits[0][joint_ids]
+                    self._solvers[name].update_with_robot_limit(joint_limits)
```
yuecideng, line 939: We should check whether the solver cfg has configured with user qpos limit. If it is specify, we should not call this method to update.

yuecideng (changes requested):

yuecideng (approved):

PR #289, opened 2026-06-03
Update grasp pose generator and pick up action

matafela:
# Description
Upgrade grasp pose generator and pick up action.

TODO:
- [x] grasp pose nms
- [x] more grasp pose sampler from different approach direction
- [x] return top k but not one grasp pose. 
- [x] pick up action choose ik valid grasp pose

## Type of change

## Checklist


embodichain/lab/sim/atomic_actions/actions.py
```diff
@@ -470,10 +470,43 @@ def _resolve_grasp_pose(
...
             obj_poses=obj_poses, approach_direction=self.approach_direction
         )
-        return is_success, grasp_xpos, open_length
+
+        # Get best grasp pose for each object
+        n_envs = obj_poses.shape[0]
+        init_qpos = self.robot.get_qpos(name=self.cfg.control_part)
+        grasp_xpos_list = []
+        is_success_list = []
+        for i in range(n_envs):
```
yuecideng: This also can be vectorized.

embodichain/utils/math.py
```diff
@@ -2265,3 +2265,114 @@ def get_offset_pose(
         return pose_to_change
 
     return offset_pose
+
+
+def pose_nms_indices(
```
yuecideng, line 2303: Add unitest for this function. Can we combine the two functions?

embodichain/toolkits/graspkit/pg_grasp/antipodal_generator.py
```diff
@@ -73,12 +73,18 @@ class GraspGeneratorCfg:
     number of sampled surface points, ray perturbation angle, and gripper jaw
     distance limits. See :class:`AntipodalSamplerCfg` for details."""
 
-    max_deviation_angle: float = np.pi / 6
```
yuecideng, line 76: Add a benchmark for the updated grasp generation algorithm. Two dimensions, 1. number of vertices. 2. number of grasp samples.

yuecideng (changes requested):

embodichain/utils/math.py
matafela: Pose nms was removed currently.

embodichain/toolkits/graspkit/pg_grasp/antipodal_generator.py
matafela: Add benchmark for surface sample point number and selected top grasp pose number.

PR #306, opened 2026-06-15
Add action-agent runtime and generated config pipeline

skywhite1024:
# Description

Add the action-agent runtime and generated-config pipeline for image-to-tabletop manipulation demos.

This PR introduces the reviewable vertical slice for running generated simulation tasks end to end:

- `TaskAgent` and deterministic `CompileAgent` wrappers for nominal task-graph generation, validation, and execution.
- JSON graph loading, schema validation, and deterministic start-to-goal graph compilation.
- Atomic-action spec normalization and execution bridge to `embodichain.lab.sim.atomic_actions`.
- Tableware env adapter for agent state, IK/FK, configured success checks, and action-list creation.
- `run_agent.py` CLI for executing an existing generated `gym_config` + `agent_config` pair.
- UR5 basket / relative-placement config generation from exported gym projects.
- Image2Tabletop and prompt-to-geometry service integration helpers used by the demo pipeline.
- One-command orchestration support for Demo1/Demo2/Demo3 generated configs.
- Unit tests for LLM usage tracking, graph schema validation, backend atomic-action runtime integration, and generated config behavior.

## Reviewer Cleanup

Addressed the review feedback around stale and invalid action-agent code:

- Removed unsupported `orientation` from target-pose schema and added regression coverage that rejects it.
- Removed stale `_build_action_cfg_and_start()` helper by inlining cfg/start-qpos construction.
- Kept only the runtime atomic-action implementation; tableware adapter code now imports the current env registration path.
- Rejected legacy/non-atomic action schemas in both runtime normalization and graph compilation.
- Removed monitor/recovery compatibility leftovers from the nominal runtime path.
- Removed unused compile LLM wiring while preserving existing generated config compatibility.

Fixes # N/A

## Type of change

- Cleanup/refactor in response to review feedback

## Screenshots

N/A. This PR adds runtime, config generation, and CLI infrastructure without UI changes.

## Validation

```bash
conda run -n embodichain040 black \
  embodichain/gen_sim/action_agent_pipeline/agents/compile_agent.py \
  embodichain/gen_sim/action_agent_pipeline/agents/llm.py \
  embodichain/gen_sim/action_agent_pipeline/cli/run_agent.py \
  embodichain/gen_sim/action_agent_pipeline/env_adapters/tableware/base_agent_env.py \
  embodichain/gen_sim/action_agent_pipeline/generation/prompt_builders.py \
  embodichain/gen_sim/action_agent_pipeline/runtime/atom_actions.py \
  embodichain/gen_sim/action_agent_pipeline/runtime/graph_compiler.py \
  tests/gen_sim/action_agent_pipeline/test_backend_atomic_runtime.py \
  tests/gen_sim/action_agent_pipeline/test_demo3_semantic_grasp_integration.py \
  tests/gen_sim/action_agent_pipeline/test_graph_spec_backend_atomic.py
```

Result: all listed files were already formatted.

```bash
conda run -n embodichain040 python -m pytest \
  tests/gen_sim/action_agent_pipeline/test_backend_atomic_runtime.py \
  tests/gen_sim/action_agent_pipeline/test_graph_spec_backend_atomic.py \
  tests/gen_sim/action_agent_pipeline/test_llm_usage.py \
  -q
```

Result: `20 passed`.

Additional smoke checks:

- Existing generated config compatibility for `CompileAgent(prompt_name="compile_agent_graph", ...)`.
- `orientation` is rejected as an unsupported target-pose field.

## Checklist

- [x] I have run formatting on changed files.
- [x] Dependencies have been updated, if applicable. No dependency update is required.
- [x] Description updated to match the current PR scope and reviewer cleanup.


XuanchaoPENG:
1. use_place_action is the invalid param.
2. duplicate names of atomic actions under `action_agent_pipeline/env_adapters/tableware` and `action_agent_pipeline/runtime`. only keep one implement of atomic actions.
3. invalid code like `sync_agent_state_from_robot`, `extract_drive_calls`, etc.
4. the action edge contains the non-atomic action.
5. invalid functions like `_build_action_cfg_and_start()` in `action_agent_pipeline/runtime/atom_actions.py`.
6. `orientation` seems like no use in prompt.

skywhite1024:
Thanks for the detailed review. I updated the PR branch and addressed the action-agent cleanup items:

- Removed `use_place_action` and verified it no longer appears in `action_agent_pipeline`.
- Kept only the runtime atomic-action implementation; the tableware adapter now imports the current env registration path, and the Demo3 semantic grasp integration test imports `agent_env` instead of the removed `tableware.atomic_actions` module.
- Removed stale/invalid leftovers including `_build_action_cfg_and_start`, monitor return fields, monitor runtime args, and unused compile LLM wiring.
- The graph compiler and runtime now reject legacy/non-atomic action schemas (`action`, `fn`/`kwargs`, unsupported edge fields).
- Removed `orientation` from the supported target-pose schema and added regression coverage that rejects it as an unsupported field.

Validation in `embodichain040`:

```bash
conda run -n embodichain040 python -m pytest \
  tests/gen_sim/action_agent_pipeline/test_backend_atomic_runtime.py \
  tests/gen_sim/action_agent_pipeline/test_graph_spec_backend_atomic.py \
  tests/gen_sim/action_agent_pipeline/test_llm_usage.py \
  -q
```

Result: `20 passed`.

I also updated the PR description because the PR branch has now been synced to the current local branch, so the scope is broader than the original runtime-only slice and includes generated config / demo pipeline support.

embodichain/gen_sim/action_agent_pipeline/cli/generate_action_agent_config.py
```diff
@@ -0,0 +1,246 @@
+# ----------------------------------------------------------------------------
```
yuecideng, line 1: Rrename this file to a more general name.

yuecideng:
### Code review

Branch: 17 commits, 57 files, +17,269 LOC. The PR adds the action-agent runtime + generated-config pipeline for image-to-tabletop manipulation. The cleanup items from the prior review (duplicate atomic-actions, `use_place_action`, stale helpers) were verified as resolved.

#### Good design patterns

1. **Clean three-layer separation in `runtime/`** — `graph_compiler.py` (JSON→graph), `task_graph.py` (pure graph + linear traversal), `atom_actions.py` (execution) each own one concern; the compiler injects `graph_cls`/`action_module` so layers are independently testable.
2. **Single schema boundary** — `AtomicActionSpec` (frozen dataclass) + `normalize_atomic_action_spec` is the one validation entry; legacy schemas (`fn`, `action`, `target`) are rejected with actionable messages.
3. **Compile-time path validation** — `_validate_nominal_path` rejects duplicate node/edge ids, dangling refs, cycles, and unreachable goals before runtime, so `AgentTaskGraph.run()` can assume a clean linear path.
4. **Compiled-graph cache with hash + schema version** — SHA-256 of canonical sorted-key JSON + `COMPILED_GRAPH_SCHEMA_VERSION`; stale artifacts self-heal.
5. **Content-addressed CoACD mesh cache** — keys cache on `source_sha256 + policy_version + dexsim_engine_version + transform`, so policy bumps auto-invalidate.
6. **Consistent HTTP error envelope** across SAM3/SAM3D/LLM clients — explicit `timeout_s`, typed errors, JSON-shape validation.
7. **Zip-slip protection** in `_safe_extract_zip`.
8. **Declarative compositional success evaluator** — recursive `all`/`any`/`not` over a leaf-predicate catalog.
9. **Non-invasive LLM usage tracking** — thin proxy over LangChain, process-local env vars, scrubbed from subprocess envs.
10. **Network-gated tests** — `pytest.mark.skipif(... RUN_DEXSIM_GRASP_TESTS ...)` and `pytest.importorskip(...)` keep the suite green without services.

#### Points to improve

**Correctness / robustness**

1. **Schema/sentinel mismatch in `graph_compiler.py`** — `_validate_task_spec` only treats JSON `null` as empty, but `_compile_action` also coerces `""`/`"none"`/`"null"`. An edge like `{"left_arm_action": "", "right_arm_action": "none"}` passes validation, compiles to both-None, and fails later in `execute_parallel_atomic_actions` with a worse error.

https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/runtime/graph_compiler.py#L156-L160

2. **`success.py::_pose()` crashes opaquely on bad uid** — `env.sim.get_rigid_object(uid)` returns `None` for a typoed object id, then `.get_local_pose(...)` raises `AttributeError`. Every leaf predicate funnels through `_pose`. Add an explicit None check raising `ValueError(f"Unknown rigid object uid: {uid!r}")`.

https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/env_adapters/tableware/success.py#L100-L103

3. **Non-atomic cache writes** — `coacd_cache_bridge.py` (`pickle.dump` to final path) and `coacd_cache.py` (CoACD `.obj` write). An interrupt leaves a corrupt file that future readers silently consume. Use a sibling temp + `os.replace()`.

https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/runtime/coacd_cache_bridge.py#L174-L183

4. **Unbounded retries / polls** — `prompt2geometry/pipeline.py` (`while True` with bare `except Exception` on JSON extraction) and `image2tabletop_client.py::wait_for_job` (polls forever with no deadline). Add max-attempts + backoff + timeout.

https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/gym_project_api/prompt2geometry/pipeline.py#L334-L342

5. **Silent exception swallowing** — `atom_actions.py:957` (`except Exception: return` on grasp-collision cache prep — bugs get masked, collision checking silently degrades); `agents/llm.py::_create_llm_safe` returns `None` with no log — root cause lost. Narrow clauses and log.

https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/runtime/atom_actions.py#L956-L958

6. **`resolve_arm_side` doesn't raise** — calls `log_error(error_type=ValueError)` but `log_error` only logs; the invalid side is returned and downstream `_select_arm_parts` produces confusing shape errors.

https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/runtime/atom_action_utils.py#L40-L48

7. **`task_agent.py` cache has no input hash** — unlike `CompileAgent`, a stale `agent_task_graph.json` is loaded silently when prompt/task_name changes but the file path doesn't. Add a prompt-version or input-hash check.

https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/agents/task_agent.py#L55-L59

8. **`env=None` defaults on required params** — `execute_parallel_atomic_actions` and `AgentTaskGraph.run`. Env is dereferenced immediately; None produces an opaque `AttributeError`. Make required.

https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/runtime/atom_actions.py#L436-L440

9. **`utils/mllm.py::apply_proxy_env` mutates `os.environ` globally** as a side effect of `create_openai_client`/`create_chat_openai` — leaks across tests and parallel workers. Pass `http_client=`/proxies to constructors instead.

https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/utils/mllm.py#L39-L46

**Structure / contracts**

10. **`ur5_basket_config.py` is 3,665 lines** holding ≥6 separable concerns (scene parsing, target replacement, LLM role refinement, relative-placement geometry, GLB/glTF parsing, static robot/sensor/light templates). At minimum split out a `glb_io.py` (note: `_read_glb` in this file duplicates `mesh_frame_normalization.py::_read_glb`) and move static config templates to YAML/JSON data files.

https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/generation/ur5_basket_config.py

11. **`run_agent_pipeline.py` is 1,329 lines** mixing arg parsing, two subprocess stages, rigid-object alias resolution, history lookup, and manifest writing. Split: target-replacement resolution, image2scene stage, arg parser + `_DEFAULT_*` constants into separate modules.

https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/cli/run_agent_pipeline.py

12. **`base_agent_env.py::_init_agents` fragile to reserved keys** — `_agent_config_with_prompt_keys` strips `prompt_kwargs` but not `task_name`/`config_dir`; an `Agent` config block with either key raises `TypeError: got multiple values for keyword argument`.

https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/env_adapters/tableware/base_agent_env.py#L47-L61

13. **`agent_env.py::__init__` double-consumes `**kwargs`** — forwards to both `EmbodiedEnv.__init__` and `_init_agents`. The contract is fragile and undocumented.

https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/env_adapters/tableware/agent_env.py#L36-L42

14. **`execute_parallel_atomic_actions` couples build + step + UI** — root of the previously-flagged "executes immediately" concern. Split into `build_parallel_action_stream(...)` (pure) and `step_env_with_actions(env, actions)`.

https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/runtime/atom_actions.py#L487-L491

15. **Redundant double-normalization** — `_compile_action` normalizes via `normalize_atomic_action_spec` and stores a dict; `execute_atomic_action` re-normalizes on every execution via `AtomicActionSpec.from_mapping`. Cache the `AtomicActionSpec` at compile time.

16. **Hardcoded intranet IPs as defaults** — `192.168.3.23:{5013,5015,5016,4523}` in `pipeline.py`, `image2tabletop_client.py`, `zimage_client.py`, `run_agent_pipeline.py`. Make required or env-only.

17. **`utils/__init__.py` is empty and missing `__all__`** — AGENTS.md requires `__all__` in every public module. All other `__init__.py` files are populated correctly; every source file has the Apache 2.0 header and `from __future__ import annotations`.

18. **Inconsistent timestamps** — pipeline records use local time / second precision; usage logs use UTC / millisecond precision. Pick one convention.

19. **`test_backend_atomic_runtime.py::_FakeBackendAction.capture` is a class-level mutable** shared across instances. Reset via fixture or make it instance-scoped.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

embodichain/gen_sim/action_agent_pipeline/env_adapters/tableware/agent_env.py
```diff
@@ -0,0 +1,54 @@
...
+from embodichain.gen_sim.action_agent_pipeline.env_adapters.tableware.success import (
+    evaluate_configured_success,
+)
+from embodichain.lab.gym.utils.registration import register_env
+
+__all__ = ["AtomicActionsAgentEnv"]
+
+
+@register_env("AtomicActionsAgent-v3", max_episode_steps=600)
+class AtomicActionsAgentEnv(BaseAgentEnv, EmbodiedEnv):
```
yuecideng: I thinl you should refactor the old agent env implementation. We can directly inherit the EmbodiedEnv and add agentic related APIs , member to the AgenticGenSimEnv. 
Also change the name to `AgenticGenSimEnv` would be better.

yuecideng:
### Module structure: `agents/hierarchy/` vs `gen_sim/action_agent_pipeline/agents/`

The two modules are **partially redundant** — not duplicates of the same feature, but an in-progress migration from "LLM generates Python code that is `exec()`'d" (old `agents/hierarchy/`) to "LLM emits a JSON task graph that is compiled and executed as atomic actions" (new `gen_sim/.../agents/`).

#### Redundancy map

| File | `agents/hierarchy/` | `gen_sim/.../agents/` | Verdict |
|---|---|---|---|
| `agent_base.py` | `AgentBase` ABC + `_resolve_prompt_path` (95 lines) | `AgentBase` ABC + `_resolve_prompt_path` (95 lines) | Near-verbatim duplicate. Only diffs: gen_sim adds `from __future__ import annotations` + PEP 604 union, and `get_composed_observations()` drops the `env.get_obs_for_agent()` coupling. `_resolve_prompt_path` is byte-identical. |
| `task_agent.py` | Emits free-text plan → `agent_generated_plan.txt` | Emits JSON task graph → `agent_task_graph.json` | Same skeleton (constructor, `generate()` flow, `act()` pass-through), different output contract. |
| `llm.py` | Azure factory + 4 module-level instances | OpenAI factory via shared `utils/mllm.py` + 1 instance | Redundant factory. Gen_sim already factored the shared factory out to `utils/mllm.py`; hierarchy still hardcodes Azure. |
| `code_agent.py` | LLM writes Python → `exec()` with AST kwargs injection (289 lines) | — | Unique to old design; replaced by `CompileAgent` + `graph_compiler`. |
| `validation_agent.py` | Vision-LLM execution validator + view selector (241 lines) | — | Unique to old design; new pipeline uses declarative `env_adapters/.../success.py` predicates instead. |

#### External usage

- `agents/hierarchy/` is imported only from `lab/gym/envs/tasks/tableware/base_agent_env.py:23-26` (old tableware task).
- `gen_sim/.../agents/` is imported only from `gen_sim/action_agent_pipeline/env_adapters/tableware/base_agent_env.py:30-36` (new gen_sim path).

There's a parallel old/new split at the env-adapter layer too. Note `agents/hierarchy/code_agent.py:216` imports `embodichain.lab.sim.atom_actions` (singular, old) while the new pipeline uses `embodichain.lab.sim.atomic_actions/` (plural package, new) — same generational split exists at the sim layer.

#### Improvement options

1. **Finish the migration, delete `agents/hierarchy/`.** Migrate the one consumer (`lab/gym/envs/tasks/tableware/base_agent_env.py`) to the gen_sim pipeline, then remove `agents/hierarchy/` entirely. `ValidationAgent` (vision-LLM success checking) has no replacement, so either port it to `env_adapters/.../success.py` or accept losing that capability.
2. **Extract shared base to `agents/hierarchy/_base.py`, have both pipelines inherit.** Removes the `AgentBase` + `_resolve_prompt_path` + LLM-factory duplication. Tradeoff: introduces cross-package coupling (gen_sim would depend on `agents/hierarchy/` or vice versa); only worth it if Option 1 is blocked.
3. **Leave as-is, document the split.** Add a deprecation docstring at the top of `agents/hierarchy/` marking it as superseded-by-gen_sim.

**Recommendation:** Option 1 if the old tableware task is still actively used; Option 3 as a holding pattern if it's already abandoned. Option 2 is a trap — it locks in the duplication by making it structural rather than temporary.

The real question for the PR author: is the old `code_agent` (LLM-writes-Python) path still needed for anything the JSON-graph path can't do yet? If not, finish the migration in a follow-up and delete `agents/hierarchy/`.

🤖 Generated with [Claude Code](https://claude.ai/code)

embodichain/gen_sim/action_agent_pipeline/cli/generate_action_agent_config.py
skywhite1024: Done. I renamed the UR5/basket-specific entry point to the more general `generate_action_agent_config.py` and moved the generation implementation to `generation/action_agent_config.py`.
The old `generate_ur5_basket_config.py` / `generation/ur5_basket_config.py` files were removed, and the tests now import the generalized API.

skywhite1024:
> ### Code review
> Branch: 17 commits, 57 files, +17,269 LOC. The PR adds the action-agent runtime + generated-config pipeline for image-to-tabletop manipulation. The cleanup items from the prior review (duplicate atomic-actions, `use_place_action`, stale helpers) were verified as resolved.
> 
> #### Good design patterns
> 1. **Clean three-layer separation in `runtime/`** — `graph_compiler.py` (JSON→graph), `task_graph.py` (pure graph + linear traversal), `atom_actions.py` (execution) each own one concern; the compiler injects `graph_cls`/`action_module` so layers are independently testable.
> 2. **Single schema boundary** — `AtomicActionSpec` (frozen dataclass) + `normalize_atomic_action_spec` is the one validation entry; legacy schemas (`fn`, `action`, `target`) are rejected with actionable messages.
> 3. **Compile-time path validation** — `_validate_nominal_path` rejects duplicate node/edge ids, dangling refs, cycles, and unreachable goals before runtime, so `AgentTaskGraph.run()` can assume a clean linear path.
> 4. **Compiled-graph cache with hash + schema version** — SHA-256 of canonical sorted-key JSON + `COMPILED_GRAPH_SCHEMA_VERSION`; stale artifacts self-heal.
> 5. **Content-addressed CoACD mesh cache** — keys cache on `source_sha256 + policy_version + dexsim_engine_version + transform`, so policy bumps auto-invalidate.
> 6. **Consistent HTTP error envelope** across SAM3/SAM3D/LLM clients — explicit `timeout_s`, typed errors, JSON-shape validation.
> 7. **Zip-slip protection** in `_safe_extract_zip`.
> 8. **Declarative compositional success evaluator** — recursive `all`/`any`/`not` over a leaf-predicate catalog.
> 9. **Non-invasive LLM usage tracking** — thin proxy over LangChain, process-local env vars, scrubbed from subprocess envs.
> 10. **Network-gated tests** — `pytest.mark.skipif(... RUN_DEXSIM_GRASP_TESTS ...)` and `pytest.importorskip(...)` keep the suite green without services.
> 
> #### Points to improve
> **Correctness / robustness**
> 
> 1. **Schema/sentinel mismatch in `graph_compiler.py`** — `_validate_task_spec` only treats JSON `null` as empty, but `_compile_action` also coerces `""`/`"none"`/`"null"`. An edge like `{"left_arm_action": "", "right_arm_action": "none"}` passes validation, compiles to both-None, and fails later in `execute_parallel_atomic_actions` with a worse error.
> 
> https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/runtime/graph_compiler.py#L156-L160
> 
> 2. **`success.py::_pose()` crashes opaquely on bad uid** — `env.sim.get_rigid_object(uid)` returns `None` for a typoed object id, then `.get_local_pose(...)` raises `AttributeError`. Every leaf predicate funnels through `_pose`. Add an explicit None check raising `ValueError(f"Unknown rigid object uid: {uid!r}")`.
> 
> https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/env_adapters/tableware/success.py#L100-L103
> 
> 3. **Non-atomic cache writes** — `coacd_cache_bridge.py` (`pickle.dump` to final path) and `coacd_cache.py` (CoACD `.obj` write). An interrupt leaves a corrupt file that future readers silently consume. Use a sibling temp + `os.replace()`.
> 
> https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/runtime/coacd_cache_bridge.py#L174-L183
> 
> 4. **Unbounded retries / polls** — `prompt2geometry/pipeline.py` (`while True` with bare `except Exception` on JSON extraction) and `image2tabletop_client.py::wait_for_job` (polls forever with no deadline). Add max-attempts + backoff + timeout.
> 
> https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/gym_project_api/prompt2geometry/pipeline.py#L334-L342
> 
> 5. **Silent exception swallowing** — `atom_actions.py:957` (`except Exception: return` on grasp-collision cache prep — bugs get masked, collision checking silently degrades); `agents/llm.py::_create_llm_safe` returns `None` with no log — root cause lost. Narrow clauses and log.
> 
> https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/runtime/atom_actions.py#L956-L958
> 
> 6. **`resolve_arm_side` doesn't raise** — calls `log_error(error_type=ValueError)` but `log_error` only logs; the invalid side is returned and downstream `_select_arm_parts` produces confusing shape errors.
> 
> https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/runtime/atom_action_utils.py#L40-L48
> 
> 7. **`task_agent.py` cache has no input hash** — unlike `CompileAgent`, a stale `agent_task_graph.json` is loaded silently when prompt/task_name changes but the file path doesn't. Add a prompt-version or input-hash check.
> 
> https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/agents/task_agent.py#L55-L59
> 
> 8. **`env=None` defaults on required params** — `execute_parallel_atomic_actions` and `AgentTaskGraph.run`. Env is dereferenced immediately; None produces an opaque `AttributeError`. Make required.
> 
> https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/runtime/atom_actions.py#L436-L440
> 
> 9. **`utils/mllm.py::apply_proxy_env` mutates `os.environ` globally** as a side effect of `create_openai_client`/`create_chat_openai` — leaks across tests and parallel workers. Pass `http_client=`/proxies to constructors instead.
> 
> https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/utils/mllm.py#L39-L46
> 
> **Structure / contracts**
> 
> 10. **`ur5_basket_config.py` is 3,665 lines** holding ≥6 separable concerns (scene parsing, target replacement, LLM role refinement, relative-placement geometry, GLB/glTF parsing, static robot/sensor/light templates). At minimum split out a `glb_io.py` (note: `_read_glb` in this file duplicates `mesh_frame_normalization.py::_read_glb`) and move static config templates to YAML/JSON data files.
> 
> https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/generation/ur5_basket_config.py
> 
> 11. **`run_agent_pipeline.py` is 1,329 lines** mixing arg parsing, two subprocess stages, rigid-object alias resolution, history lookup, and manifest writing. Split: target-replacement resolution, image2scene stage, arg parser + `_DEFAULT_*` constants into separate modules.
> 
> https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/cli/run_agent_pipeline.py
> 
> 12. **`base_agent_env.py::_init_agents` fragile to reserved keys** — `_agent_config_with_prompt_keys` strips `prompt_kwargs` but not `task_name`/`config_dir`; an `Agent` config block with either key raises `TypeError: got multiple values for keyword argument`.
> 
> https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/env_adapters/tableware/base_agent_env.py#L47-L61
> 
> 13. **`agent_env.py::__init__` double-consumes `**kwargs`** — forwards to both `EmbodiedEnv.__init__` and `_init_agents`. The contract is fragile and undocumented.
> 
> https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/env_adapters/tableware/agent_env.py#L36-L42
> 
> 14. **`execute_parallel_atomic_actions` couples build + step + UI** — root of the previously-flagged "executes immediately" concern. Split into `build_parallel_action_stream(...)` (pure) and `step_env_with_actions(env, actions)`.
> 
> https://github.com/DexForce/EmbodiChain/blob/24901bddef40c7fe69cae4e79d4d235aa4d73462/embodichain/gen_sim/action_agent_pipeline/runtime/atom_actions.py#L487-L491
> 
> 15. **Redundant double-normalization** — `_compile_action` normalizes via `normalize_atomic_action_spec` and stores a dict; `execute_atomic_action` re-normalizes on every execution via `AtomicActionSpec.from_mapping`. Cache the `AtomicActionSpec` at compile time.
> 16. **Hardcoded intranet IPs as defaults** — `192.168.3.23:{5013,5015,5016,4523}` in `pipeline.py`, `image2tabletop_client.py`, `zimage_client.py`, `run_agent_pipeline.py`. Make required or env-only.
> 17. **`utils/__init__.py` is empty and missing `__all__`** — AGENTS.md requires `__all__` in every public module. All other `__init__.py` files are populated correctly; every source file has the Apache 2.0 header and `from __future__ import annotations`.
> 18. **Inconsistent timestamps** — pipeline records use local time / second precision; usage logs use UTC / millisecond precision. Pick one convention.
> 19. **`test_backend_atomic_runtime.py::_FakeBackendAction.capture` is a class-level mutable** shared across instances. Reset via fixture or make it instance-scoped.
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

Thanks for the thorough review. I addressed the cleanup items in the latest update:
- Aligned graph validation with runtime action normalization, so `""`, `"none"`, and `"null"` are rejected at compile time when both arms are empty.
- Added an explicit unknown-object check in `success._pose()`.
- Switched CoACD/grasp cache writes to temp-file + `os.replace()`.
- Added bounded retry/timeout behavior for prompt2geometry naming and Image2Tabletop polling.
- Narrowed/logged grasp-cache fallback exceptions and logged LLM initialization failures.
- Made `resolve_arm_side()` raise clear `ValueError`s.
- Added prompt-hash metadata for `TaskAgent` cache validation.
- Made `env` required for graph/action execution paths.
- Changed proxy handling to per-client `httpx.Client` instead of mutating `os.environ`.
- Split the large generation and CLI modules into focused files, including `glb_io.py`, config/template helpers, target replacement, image2scene stage, parser/defaults, and pipeline runner modules.
- Split action stream building from environment stepping via `build_parallel_action_stream()` and `step_env_with_actions()`.
- Compile-time action normalization now stores `AtomicActionSpec` objects to avoid redundant normalization.
- Removed hardcoded intranet service defaults and kept service URLs env/argument driven.
- Added `__all__` and aligned timestamp records to UTC milliseconds.
- Reset the fake backend capture through an autouse fixture.

embodichain/gen_sim/action_agent_pipeline/env_adapters/tableware/agent_env.py
skywhite1024: Done. I refactored the new action-agent env to `AgenticGenSimEnv`, and it now directly inherits `EmbodiedEnv`.
The old `BaseAgentEnv` layer in the action-agent pipeline was removed. Agent-specific kwargs are split before calling `EmbodiedEnv.__init__`, and reserved config keys such as `task_name` / `config_dir` are explicitly rejected to avoid duplicate keyword issues.
I kept `AtomicActionsAgentEnv = AgenticGenSimEnv` as a compatibility alias for the registered env/API surface.

skywhite1024:
> ### Module structure: `agents/hierarchy/` vs `gen_sim/action_agent_pipeline/agents/`
> The two modules are **partially redundant** — not duplicates of the same feature, but an in-progress migration from "LLM generates Python code that is `exec()`'d" (old `agents/hierarchy/`) to "LLM emits a JSON task graph that is compiled and executed as atomic actions" (new `gen_sim/.../agents/`).
> 
> #### Redundancy map
> File	`agents/hierarchy/`	`gen_sim/.../agents/`	Verdict
> `agent_base.py`	`AgentBase` ABC + `_resolve_prompt_path` (95 lines)	`AgentBase` ABC + `_resolve_prompt_path` (95 lines)	Near-verbatim duplicate. Only diffs: gen_sim adds `from __future__ import annotations` + PEP 604 union, and `get_composed_observations()` drops the `env.get_obs_for_agent()` coupling. `_resolve_prompt_path` is byte-identical.
> `task_agent.py`	Emits free-text plan → `agent_generated_plan.txt`	Emits JSON task graph → `agent_task_graph.json`	Same skeleton (constructor, `generate()` flow, `act()` pass-through), different output contract.
> `llm.py`	Azure factory + 4 module-level instances	OpenAI factory via shared `utils/mllm.py` + 1 instance	Redundant factory. Gen_sim already factored the shared factory out to `utils/mllm.py`; hierarchy still hardcodes Azure.
> `code_agent.py`	LLM writes Python → `exec()` with AST kwargs injection (289 lines)	—	Unique to old design; replaced by `CompileAgent` + `graph_compiler`.
> `validation_agent.py`	Vision-LLM execution validator + view selector (241 lines)	—	Unique to old design; new pipeline uses declarative `env_adapters/.../success.py` predicates instead.
> #### External usage
> * `agents/hierarchy/` is imported only from `lab/gym/envs/tasks/tableware/base_agent_env.py:23-26` (old tableware task).
> * `gen_sim/.../agents/` is imported only from `gen_sim/action_agent_pipeline/env_adapters/tableware/base_agent_env.py:30-36` (new gen_sim path).
> 
> There's a parallel old/new split at the env-adapter layer too. Note `agents/hierarchy/code_agent.py:216` imports `embodichain.lab.sim.atom_actions` (singular, old) while the new pipeline uses `embodichain.lab.sim.atomic_actions/` (plural package, new) — same generational split exists at the sim layer.
> 
> #### Improvement options
> 1. **Finish the migration, delete `agents/hierarchy/`.** Migrate the one consumer (`lab/gym/envs/tasks/tableware/base_agent_env.py`) to the gen_sim pipeline, then remove `agents/hierarchy/` entirely. `ValidationAgent` (vision-LLM success checking) has no replacement, so either port it to `env_adapters/.../success.py` or accept losing that capability.
> 2. **Extract shared base to `agents/hierarchy/_base.py`, have both pipelines inherit.** Removes the `AgentBase` + `_resolve_prompt_path` + LLM-factory duplication. Tradeoff: introduces cross-package coupling (gen_sim would depend on `agents/hierarchy/` or vice versa); only worth it if Option 1 is blocked.
> 3. **Leave as-is, document the split.** Add a deprecation docstring at the top of `agents/hierarchy/` marking it as superseded-by-gen_sim.
> 
> **Recommendation:** Option 1 if the old tableware task is still actively used; Option 3 as a holding pattern if it's already abandoned. Option 2 is a trap — it locks in the duplication by making it structural rather than temporary.
> 
> The real question for the PR author: is the old `code_agent` (LLM-writes-Python) path still needed for anything the JSON-graph path can't do yet? If not, finish the migration in a follow-up and delete `agents/hierarchy/`.
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)

Thanks for calling this out. I finished the migration and removed the legacy Python-code-generation action-agent path.
What changed:
- Deleted `embodichain/agents/hierarchy/` and the old prompt files used only by that path.
- Removed the legacy singular `embodichain.lab.sim.atom_actions` module.
- Removed the old `PourWaterAgent-v3` and `RearrangementAgent-v3` registrations that depended on `BaseAgentEnv`.
- Kept the non-agent task envs (`PourWater-v3`, `Rearrangement-v3`) intact.
- Updated API docs so Sphinx no longer imports the removed legacy modules.
The supported demos now go only through `embodichain.gen_sim.action_agent_pipeline` and `AtomicActionsAgent-v3`.

embodichain/gen_sim/action_agent_pipeline/agents/compile_agent.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class CompileAgent(AgentBase):
+    """Compile and execute nominal atomic-action graph specs."""
+
+    query_prefix = "# "
+    query_suffix = "."
+    prompt_kwargs: dict[str, dict[str, Any]]
+
+    def __init__(self, **kwargs) -> None:
```
yuecideng, line 43: CompileAgent.__init__ bypasses AgentBase.__init__, so config_dir resolution and prompt preloading are skipped. Start with `super().__init__(**kwargs)` and remove the duplicated `setattr` loop.

embodichain/gen_sim/action_agent_pipeline/agents/compile_agent.py
```diff
@@ -0,0 +1,119 @@
...
+            compile_agent_graph_from_file,
+        )
+
+        runtime_kwargs = _runtime_kwargs(kwargs, getattr(self, "prompt_kwargs", {}))
+        graph = compile_agent_graph_from_file(graph_file_path)
+        result = graph.run(**runtime_kwargs)
+        print("Compiled agent graph executed successfully.")
+        return result
+
+    def get_composed_observations(self, **kwargs):
```
yuecideng, line 96: This overrides `get_composed_observations` with different semantics (`dict(kwargs)` instead of merging prompt contents), breaking Liskov substitution. Align with `AgentBase` or rename the method.

embodichain/gen_sim/action_agent_pipeline/gym_project_api/prompt2geometry/dimensions.py
```diff
@@ -0,0 +1,127 @@
...
+- Keep reason short and specific.
+</notes>
+""".strip()
+
+
+def estimate_real_dimensions(
+    *,
+    object_prompt: str,
+    client: OpenAICompatibleClient,
+    max_attempts: int | None = None,
```
yuecideng, line 74: `max_attempts=None` makes the retry loop infinite when the LLM repeatedly returns invalid JSON. Set a finite default (e.g., 3–5) or require callers to pass one.

embodichain/gen_sim/action_agent_pipeline/gym_project_api/prompt2geometry/pipeline.py
```diff
@@ -0,0 +1,623 @@
...
+    return OpenAICompatibleClient(
+        api_key=str(api_key),
+        model=str(model),
+        base_url=str(base_url),
+        timeout_s=request.llm_timeout_s,
+        usage_stage=f"prompt2geometry.{purpose}",
+    )
+
+
+def _cleanup_output_root(output_root: Path, *, keep_path: Path | None) -> None:
```
yuecideng, line 424: `_cleanup_output_root` can delete arbitrary directory contents and follows directory symlinks outside `output_root`. Add path validation (ensure it's a dedicated subdirectory), skip symlinks whose targets are outside `output_root`, and consider requiring `--allow-cleanup`.

embodichain/gen_sim/action_agent_pipeline/env_adapters/tableware/agent_env.py
```diff
@@ -0,0 +1,401 @@
...
+_REQUIRED_AGENT_KWARGS = frozenset({"agent_config", "task_name"})
+_OPTIONAL_AGENT_KWARGS = frozenset({"agent_config_path"})
+_AGENT_KWARGS = _REQUIRED_AGENT_KWARGS | _OPTIONAL_AGENT_KWARGS
+
+
+@register_env("AtomicActionsAgent-v3", max_episode_steps=600)
+class AgenticGenSimEnv(EmbodiedEnv):
+    """Config-driven agent environment for atomic-action tasks."""
+
+    def __init__(self, cfg: EmbodiedEnvCfg = None, **kwargs):
```
yuecideng, line 43: `cfg: EmbodiedEnvCfg = None` is not type-safe. Use `cfg: EmbodiedEnvCfg | None = None`, and type `reset` similarly (`options: dict[str, Any] | None = None`) with an explicit return type.

embodichain/gen_sim/action_agent_pipeline/env_adapters/tableware/agent_env.py
```diff
@@ -0,0 +1,401 @@
...
+
+    def is_task_success(self, **kwargs) -> torch.Tensor:
+        return evaluate_configured_success(self)
+
+    def compute_task_state(self, **kwargs) -> tuple[torch.Tensor, torch.Tensor, dict]:
+        success = self.is_task_success()
+        fail = torch.zeros_like(success)
+        return success, fail, {}
+
+    def _init_agents(self, agent_config, task_name, agent_config_path=None):
```
yuecideng, line 63: `_init_agents` indexes `agent_config['Agent']`, `['TaskAgent']`, `['CompileAgent']` before validating that those keys exist. A malformed config raises `KeyError` instead of the intended `ValueError`. Validate keys first, then pass the section dicts.

embodichain/gen_sim/action_agent_pipeline/env_adapters/tableware/agent_env.py
```diff
@@ -0,0 +1,401 @@
...
+    def _agent_config_with_prompt_keys(self, agent_config, allowed_keys):
+        filtered = deepcopy(agent_config)
+        prompt_kwargs = filtered.get("prompt_kwargs", {}) or {}
+        filtered["prompt_kwargs"] = {
+            key: value for key, value in prompt_kwargs.items() if key in allowed_keys
+        }
+        return filtered
+
+    def get_states(self):
+        # TODO: only support num_env = 1 for now
```
yuecideng, line 117: `get_states` calls `.squeeze(0)` on robot state, silently assuming `num_envs == 1`. Add an explicit guard that raises if `num_envs != 1`, or implement proper batched support.

embodichain/gen_sim/action_agent_pipeline/env_adapters/tableware/success.py
```diff
@@ -0,0 +1,306 @@
...
+def _object_axis_near(env, spec: Mapping[str, Any]) -> torch.Tensor:
+    object_position = _position(env, _object_name(spec))
+    axis = _axis_index(str(spec.get("axis", "y")))
+    target_value = float(spec.get("target", spec.get("value")))
+    return torch.abs(object_position[:, axis] - target_value) <= float(
+        spec.get("tolerance", 0.02)
+    )
+
+
+def _object_lifted(env, spec: Mapping[str, Any]) -> torch.Tensor:
```
yuecideng, line 222: `_object_lifted` falls back to `position[:, 2]` as the initial height when `obj_info` lacks the object. That makes `min_height` always satisfied and causes false-positive success. Raise an error or load the true initial height from the environment config.

embodichain/gen_sim/action_agent_pipeline/generation/mesh_bounds.py
```diff
@@ -0,0 +1,558 @@
...
+    component_type = int(accessor["componentType"])
+    if component_type not in _GLTF_COMPONENT_FORMATS:
+        raise ValueError(f"Unsupported GLB component type: {component_type}.")
+    component_format, component_size = _GLTF_COMPONENT_FORMATS[component_type]
+    component_count = _GLTF_TYPE_COMPONENT_COUNTS[accessor["type"]]
+    buffer_view = doc["bufferViews"][int(accessor["bufferView"])]
+    if int(buffer_view.get("buffer", 0)) != 0:
+        raise ValueError("Only GLB embedded binary buffers are supported.")
+
+    stride = int(buffer_view.get("byteStride", component_size * component_count))
```
yuecideng, line 370: `_iter_gltf_accessor_vec3` ignores the GLTF accessor `normalized` flag. For normalized integer accessors, scale component values by the normalization factor or bounding-box computations will be corrupted.

embodichain/gen_sim/action_agent_pipeline/generation/coacd_cache.py
```diff
@@ -0,0 +1,194 @@
...
+    if cache_dir is None:
+        cache_dir = _DEFAULT_CONVEX_DECOMP_DIR
+
+    mesh_md5_key = dexsim_coacd_cache_key_for_mesh(mesh_path, mesh_count=mesh_count)
+    return Path(cache_dir).expanduser().resolve() / (
+        f"{mesh_md5_key}_{int(max_convex_hull_num)}.obj"
+    )
+
+
+def dexsim_coacd_cache_key_for_mesh(
```
yuecideng, line 56: Cache key uses `hashlib.md5`, which is disallowed by many security baselines. Prefer `hashlib.sha256` (or `blake2b`) for new code.

yuecideng (changes requested):
Thanks for the refactor — the new action-agent runtime and generated-config pipeline is a solid architectural improvement. It removes arbitrary code execution, adds deterministic graph compilation, and the test coverage is good.

However, I consolidated reviews from several focused passes and there are a number of issues that should be addressed before merging. I left inline comments on the critical ones; the rest are summarized below.

## Critical (must fix)

1. **`CompileAgent` bypasses `AgentBase.__init__`** (`agents/compile_agent.py:43`) — skips `config_dir` resolution and prompt preloading.
2. **`CompileAgent.get_composed_observations` breaks Liskov substitution** (`agents/compile_agent.py:96`) — returns `dict(kwargs)` instead of merged prompt contents.
3. **Infinite retry in `estimate_real_dimensions`** (`gym_project_api/prompt2geometry/dimensions.py:74`) — `max_attempts=None` hangs on persistent LLM failures.
4. **Destructive `_cleanup_output_root`** (`gym_project_api/prompt2geometry/pipeline.py:424`) — can delete arbitrary directory contents and follow symlinks outside the target.
5. **Untyped `cfg` default in `AgenticGenSimEnv.__init__`** (`env_adapters/tableware/agent_env.py:43`) — should be `EmbodiedEnvCfg | None`.
6. **Agent config keys indexed before validation** (`env_adapters/tableware/agent_env.py:63`) — raises `KeyError` instead of `ValueError`.
7. **Silent `num_envs == 1` assumption** (`env_adapters/tableware/agent_env.py:117`) — `.squeeze(0)` will break batched envs.
8. **False-positive lifted success** (`env_adapters/tableware/success.py:222`) — missing initial height falls back to current pose.
9. **GLTF `normalized` accessor ignored** (`generation/mesh_bounds.py:370`) — corrupts bounds for quantized meshes.
10. **MD5 cache key** (`generation/coacd_cache.py:56`) — disallowed by many security baselines; use SHA-256.

## Important (should fix)

**Core / agents**
- Fix `agents/__init__.py` `__all__` to export classes (`AgentBase`, `TaskAgent`, `CompileAgent`, `create_llm`), not module names.
- Replace `assert` with explicit exceptions and mark `generate`/`act` as `@abstractmethod` in `AgentBase`.
- Add type hints to `agents/llm.py` and untyped public functions in `runtime/atom_actions.py`.
- Fix `_prepare_grasp_collision_cache_from_env_coacd` swallowing broad exceptions.
- Harden `coacd_cache_bridge` pickle cache (shared writable pickle is a code-execution vector).
- Reuse `MotionGenerator` across atomic actions instead of creating one per call.

**CLI / API**
- Fix `"None"` API-key string in `prompt2geometry/config.py` (`str(None)` becomes `"None"`).
- Fix `target_replacement1/2` help text vs. selection logic mismatch and unify `nargs` semantics between pipeline and standalone generator.
- Expose or document missing flags in `pipeline_runner.py`.
- Fix `run_agent.py` passing `int` to `action_sentence`.
- Validate URL scheme in `image2tabletop_client.py`.
- Type `pipeline_records.py` and `target_replacements.py` properly.

**Generation**
- Consolidate duplicate `_make_relative_events_config` / `_make_arrangement_events_config`.
- Extract shared arrangement/stacking bundle builder.
- Derive observation `joint_ids` from robot config (currently hard-coded DualUR5).
- Distinguish mesh load errors from missing dependencies in `_load_mesh_vertices`.
- Derive repo root from `gym_config_path`, not `Path.cwd()`.
- Unify basket bundle validator with shared UID checks.
- Skip texture extraction when normalized OBJ cache is valid.
- Use `dataclasses.replace` in `_replace_relative_spec_placements`.

**Tests**
- Enable/document opt-in simulation integration tests in CI.
- Move success-predicate tests from `test_ur5_basket_config_generation.py` to `test_tableware_success.py` and add failure cases.
- Add a real dual-arm `execute_parallel_atomic_actions` test.
- Split the ~3,500-line `test_ur5_basket_config_generation.py`.

## Conventions / minor
- Replace `print()` calls with the project logger.
- Fix the license header in `rearrangement.py`.
- Move shebangs below license headers.
- Add missing docstrings to public builder functions.

## Recommendation

Address the critical inline comments first, then the important items. Once the critical and most important issues are fixed and tests remain green, this should be ready for another review pass.

embodichain/gen_sim/action_agent_pipeline/cli/generate_action_agent_config.py
```diff
@@ -0,0 +1,245 @@
...
+        "--target_body_scale",
+        type=float,
+        default=0.7,
+        help=(
+            "Uniform body_scale for generated target objects. Basket-like "
+            "containers keep their source body_scale."
+        ),
+    )
+    parser.add_argument(
+        "--target_replacement1",
```
yuecideng, line 98: target_replacement should be refactor to support 0-N objects replacement. Also, the `target` is defined as the foregraound intractive objects

embodichain/gen_sim/action_agent_pipeline/cli/__init__.py
```diff
@@ -14,6 +14,6 @@
 # limitations under the License.
 # ----------------------------------------------------------------------------
```
yuecideng, line 16: Why put so mang files under this folder? CLI usually contains only the entry points of a functionality.

embodichain/lab/gym/envs/tasks/tableware/rearrangement.py
```diff
@@ -4,12 +4,12 @@
 # All rights reserved.
 # ----------------------------------------------------------------------------
 
-from typing import Dict, Optional
+from __future__ import annotations
```
yuecideng, line 7: Can you still have these two envs run with agent?

yuecideng (changes requested):

embodichain/gen_sim/action_agent_pipeline/generation/config_blocks.py
```diff
@@ -0,0 +1,684 @@
...
+
+
+def _make_observations_config() -> dict[str, Any]:
+    return {
+        "norm_robot_eef_joint": {
+            "func": "normalize_robot_joint_data",
+            "mode": "modify",
+            "name": "robot/qpos",
+            "params": {
+                "joint_ids": [12, 13, 14, 15],
```
yuecideng, line 274: Hardcoded `joint_ids: [12, 13, 14, 15]` is DualUR5-specific. Derive these from the robot config so the observation setup works for other arms.

embodichain/gen_sim/action_agent_pipeline/utils/llm_config.py
```diff
@@ -0,0 +1,162 @@
...
+    "DEFAULT_LLM_MODEL",
+    "ACTION_PIPELINE_LLM_ENV_PATH",
+    "GEN_CONFIG_PATH",
+    "LLM_ENV_PATH",
+    "LEGACY_LLM_ENV_PATH",
+    "SIMREADY_LLM_ENV_PATH",
+    "get_openai_compatible_llm_config",
+]
+
+DEFAULT_LLM_MODEL = "gpt-4o"
```
yuecideng, line 34: `DEFAULT_LLM_MODEL = "gpt-4o"` is hardcoded. Consider loading the default from a project config or environment variable, and document how users can override it.

embodichain/gen_sim/action_agent_pipeline/cli/pipeline_defaults.py
```diff
@@ -0,0 +1,74 @@
...
+DEFAULT_IMAGE2SCENE_ROOT = REPO_ROOT / "gym_project/environment/image2tabletop"
+DEFAULT_IMAGE2SCENE_IMAGE = "scene_image/robotwin_example.png"
+DEFAULT_IMAGE2SCENE_DOWNLOAD_DIR = "./downloads"
+DEFAULT_IMAGE2SCENE_OUTPUT_ROOT = "./generated"
+DEFAULT_IMAGE2SCENE_CONFIG = "./gen_config.json"
+DEFAULT_CONFIG_OUTPUT_DIR = DEFAULT_ACTION_AGENT_WORKSPACE / "configs/demo3_text"
+DEFAULT_PIPELINE_HISTORY = (
+    DEFAULT_ACTION_AGENT_WORKSPACE / "configs/pipeline_history.json"
+)
+DEFAULT_TASK_NAME = "Demo3_Text"
```
yuecideng, line 69: `DEFAULT_TASK_NAME = "Demo3_Text"` is demo-specific. This belongs in a demo script or CLI default, not in the library defaults module.

embodichain/gen_sim/action_agent_pipeline/cli/pipeline_defaults.py
```diff
@@ -0,0 +1,74 @@
...
+DEFAULT_TASK_TEMPLATE_NAMES = frozenset({"Demo1_Text"})
```
yuecideng, line 70: `DEFAULT_TASK_TEMPLATE_NAMES = frozenset({"Demo1_Text"})` is also demo-specific. Move to demo-level configuration.

embodichain/gen_sim/action_agent_pipeline/generation/naming.py
```diff
@@ -0,0 +1,168 @@
...
+
+def _target_runtime_suffix(base: str) -> str:
+    if base == "bread":
+        return "bread_roll"
+    return base
+
+
+def _container_runtime_uid(container: _SceneObject) -> str:
+    base = _base_name(container)
+    if "basket" in base:
```
yuecideng, line 91: Special-casing `"basket"` → `"wicker_basket"` is a hidden convention. Document it or make it configurable per task.

embodichain/gen_sim/action_agent_pipeline/generation/prompt_builders.py
```diff
@@ -0,0 +1,2073 @@
...
+    "make_basket_task_prompt",
+    "make_relative_atom_actions_prompt",
+    "make_relative_basic_background",
+    "make_relative_task_prompt",
+    "make_stacking_atom_actions_prompt",
+    "make_stacking_basic_background",
+    "make_stacking_task_prompt",
+]
+
+_BASKET_LEFT_RELEASE_OFFSET_Y = 0.04
```
yuecideng, line 41: Physical constants (`_BASKET_LEFT_RELEASE_OFFSET_Y = 0.04`, `_PLACE_LIFT_HEIGHT = 0.10`, sample intervals) are scattered as module-level literals. Consider a `@configclass` for per-task physical parameters.

embodichain/gen_sim/action_agent_pipeline/generation/relative_geometry.py
```diff
@@ -0,0 +1,532 @@
...
+def _order_inside_container_slot_indices(
+    indices: list[int],
+    *,
+    placements: Sequence[_RelativePlacementStepSpec],
+    axis: str,
+    object_configs: Mapping[str, Mapping[str, Any]],
+    container_config: Mapping[str, Any] | None,
+) -> list[int]:
+    if axis == "y":
+        side_order = {"left": 0, "right": 1}
```
yuecideng, line 429: `side_order = {"left": 0, "right": 1}` hardcodes the dual-arm convention. This should come from the robot config or arm-slot mapping.

embodichain/gen_sim/action_agent_pipeline/gym_project_api/prompt2geometry/pipeline.py
```diff
@@ -0,0 +1,623 @@
...
+class Prompt2GeometryRequest:
+    """Request for prompt-to-single-asset geometry generation."""
+
+    prompt: str
+    output_root: Path
+    target_id: str = "asset_0"
+    request_id: str = "prompt2geometry_asset_0"
+    output_name: str | None = None
+    zimage_base_url: str = ""
+    zimage_width: int = 1024
```
yuecideng, line 60: Image-generation defaults (`zimage_width=1024`, `zimage_height=1024`, `zimage_seed=42`, `zimage_num_inference_steps=8`) are hardcoded. They are overrideable via the dataclass, but the defaults should be documented or moved to a service config.

embodichain/gen_sim/action_agent_pipeline/runtime/atom_actions.py
```diff
@@ -0,0 +1,1685 @@
...
+        raise RuntimeError(
+            "Grasp annotation cache is missing and annotation is disabled; "
+            "set allow_grasp_annotation=True."
+        )
+
+    antipodal_sampler_cfg = AntipodalSamplerCfg(
+        **_cfg_supported_kwargs(
+            AntipodalSamplerCfg,
+            {
+                "n_sample": int(runtime_kwargs.get("grasp_antipodal_n_sample", 20000)),
```
yuecideng, line 1372: Grasp sampling defaults (`n_sample=20000`, `max_length=0.088`, `min_length=0.003`, `viser_port=11801`) are inline. While some are overrideable via `runtime_kwargs`, the defaults should be centralized in a config object.

embodichain/gen_sim/action_agent_pipeline/env_adapters/tableware/success.py
```diff
@@ -0,0 +1,306 @@
...
+        env,
+        str(spec.get("container", spec.get("container_uid"))),
+    )
+    xy_distance = torch.linalg.norm(
+        object_position[:, :2] - container_position[:, :2],
+        dim=-1,
+    )
+    z_offset = object_position[:, 2] - container_position[:, 2]
+    return (
+        (xy_distance <= float(spec.get("xy_radius", spec.get("radius", 0.1))))
```
yuecideng, line 163: Success tolerances (`xy_radius` default 0.1, `min_z_offset` -0.03, `max_z_offset` 0.25) are hardcoded. These should be task-configurable.

PR #321, opened 2026-06-23
Add experimental NeuralPlanner for learned waypoint pass through

yangchen73:
# Description

This PR adds an experimental `NeuralPlanner` that uses an APG transformer checkpoint to follow EEF waypoints via `MotionGenerator`.

- Add `NeuralPlanner`, `NeuralPlannerCfg`, and `NeuralPlanOptions`
- Add Franka example and unit tests
- Add docs

## Type of change

- Documentation update

## Screenshots

https://github.com/user-attachments/assets/43ef0fa3-8f78-4ae9-9854-e7ecd1f0e751

## Checklist
- [ ] Dependencies have been updated, if applicable.

### DIFF
--- a/docs/source/overview/sim/planners/neural_planner.md
+++ b/docs/source/overview/sim/planners/neural_planner.md
@@ -1,0 +1,63 @@
+# NeuralPlanner
+
+````{admonition} Experimental
+:class: warning
+
+`NeuralPlanner` is an **experimental** feature. The API, checkpoint format,
+and default parameters may change without a deprecation cycle. It is currently
+only validated on the **Franka Panda** robot.
+````
+
+`NeuralPlanner` is a learning-based EEF waypoint planner. It rolls out a
+trained APG checkpoint through `MotionGenerator` to reach Cartesian targets.
+
+## Configuration
+
+Pre-trained checkpoints are hosted on HuggingFace and can be downloaded with
+`download_neural_planner_checkpoint()` (requires `HF_TOKEN` environment variable).
+
+```python
+from embodichain.data.assets.planner_assets import download_neural_planner_checkpoint
+from embodichain.lab.sim.planners import (
+    MotionGenCfg,
+    MotionGenOptions,
+    MotionGenerator,
+    MoveType,
+    NeuralPlannerCfg,
+    PlanState,
+)
+from embodichain.lab.sim.planners.neural_planner import NeuralPlanOptions
+
+checkpoint_path = download_neural_planner_checkpoint()
+
+motion_generator = MotionGenerator(
+    cfg=MotionGenCfg(
+        planner_cfg=NeuralPlannerCfg(
+            robot_uid=robot.uid,
+            checkpoint_path=checkpoint_path,
+            control_part="main_arm",
+        )
+    )
+)
+
+result = motion_generator.generate(
+    target_states=[
+        PlanState(move_type=MoveType.EEF_MOVE, xpos=waypoint)
+        for waypoint in waypoints
+    ],
+    options=MotionGenOptions(
+        plan_opts=NeuralPlanOptions(
+            control_part="main_arm",
+            start_qpos=start_qpos,
+        ),
+    ),
+)
+```
+
+## Example
+
+```bash
+python examples/sim/planners/neural_planner.py --headless --device cuda
+```
+
+The example downloads the checkpoint automatically on first run.
--- a/embodichain/data/assets/__init__.py
+++ b/embodichain/data/assets/__init__.py
@@ -21,3 +21,5 @@ from .w1_assets import *
 from .eef_assets import *
 from .robot_assets import *
 from .scene_assets import *
+from .solver_assets import *
+from .planner_assets import *
--- a/embodichain/data/assets/planner_assets.py
+++ b/embodichain/data/assets/planner_assets.py
@@ -29,7 +29,7 @@ __all__ = ["download_neural_planner_checkpoint"]
 
 def download_neural_planner_checkpoint(
     repo_id: str = "dexforce/neural_motion_generator",
-    filename: str = "franka.pt",
+    filename: str = "franka/franka.pt",
     token: str | None = None,
     endpoint: str = _HF_ENDPOINT,
 ) -> str:
@@ -43,7 +43,7 @@ def download_neural_planner_checkpoint(
 
     Args:
         repo_id: HuggingFace repository ID.
-        filename: Checkpoint filename to download.
+        filename: Checkpoint path in the repo, e.g. ``franka/franka.pt``.
         token: HuggingFace API token. Falls back to the ``HF_TOKEN``
             environment variable or the cached token from
             ``huggingface-cli login``.
--- a/embodichain/data/assets/planner_assets.py
+++ b/embodichain/data/assets/planner_assets.py
@@ -1,0 +1,91 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2021-2026 DexForce Technology Co., Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ----------------------------------------------------------------------------
+from __future__ import annotations
+
+import os
+
+from huggingface_hub import hf_hub_download
+
+# HuggingFace endpoint. Mirrors (e.g. hf-mirror.com) often redirect to the
+# real hub without forwarding the required commit-hash response headers, so we
+# default to the canonical endpoint and rely on the system proxy when needed.
+_HF_ENDPOINT = "https://huggingface.co"
+
+__all__ = ["download_neural_planner_checkpoint"]
+
+
+def download_neural_planner_checkpoint(
+    repo_id: str = "dexforce/neural_motion_generator",
+    filename: str = "franka.pt",
+    token: str | None = None,
+    endpoint: str = _HF_ENDPOINT,
+) -> str:
+    """Download a neural planner checkpoint from HuggingFace.
+
+    The repository is gated. Either set the ``HF_TOKEN`` environment variable or
+    run ``huggingface-cli login`` before calling this function.
+
+    If your network requires an HTTP proxy, set ``HTTPS_PROXY`` or
+    ``https_proxy`` in the environment before launching Python.
+
+    Args:
+        repo_id: HuggingFace repository ID.
+        filename: Checkpoint filename to download.
+        token: HuggingFace API token. Falls back to the ``HF_TOKEN``
+            environment variable or the cached token from
+            ``huggingface-cli login``.
+        endpoint: HuggingFace-compatible endpoint URL. Defaults to
+            ``https://huggingface.co``. Mirrors that merely redirect to the
+            real hub are not supported.
+
+    Returns:
+        str: Local path to the downloaded checkpoint file.
+
+    Raises:
+        RuntimeError: If the download fails, with authentication instructions.
+    """
+    # Normalize proxy env vars: the ``requests`` library on Linux requires the
+    # lowercase form (``https_proxy``), but users typically export the uppercase
+    # form (``HTTPS_PROXY``).
+    https_proxy = os.environ.get("HTTPS_PROXY") or os.environ.get("https_proxy")
+    if https_proxy:
+        os.environ.setdefault("https_proxy", https_proxy)
+        os.environ.setdefault("HTTPS_PROXY", https_proxy)
+
+    # Allow callers to pass the token explicitly; otherwise fall back to
+    # HF_TOKEN (huggingface_hub also reads this automatically, but being
+    # explicit makes the fallback order transparent).
+    if token is None:
+        token = os.environ.get("HF_TOKEN") or None
+
+    try:
+        return hf_hub_download(
+            repo_id=repo_id,
+            filename=filename,
+            token=token,
+            endpoint=endpoint,
+        )
+    except Exception as exc:
+        raise RuntimeError(
+            f"Failed to download '{filename}' from '{repo_id}'.\n"
+            "The repository is gated and requires an authenticated HuggingFace account.\n"
+            "To fix this:\n"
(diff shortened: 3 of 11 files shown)


embodichain/lab/sim/planners/neural_planner.py
```diff
@@ -0,0 +1,489 @@
...
+        hidden_dim: int = 256,
+        transformer_nhead: int = 4,
+        transformer_num_layers: int = 2,
+        transformer_ff_dim: int | None = None,
+    ):
+        super().__init__()
+        self.num_waypoints = int(num_waypoints)
+        self.use_relative_obs = bool(use_relative_obs)
+        self.state_dim = 7 + 7 + 7 + (7 if self.use_relative_obs else 0)
+        self.waypoint_token_dim = 3 + 4 + 1 + 1
```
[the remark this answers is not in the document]
yangchen73: Added

embodichain/lab/sim/planners/neural_planner.py
```diff
@@ -0,0 +1,489 @@
...
+    def _initial_qpos(
+        self, control_part: str, start_qpos: torch.Tensor | None
+    ) -> torch.Tensor:
+        if start_qpos is None:
+            qpos = self.robot.get_qpos(name=control_part)[0]
+        else:
+            qpos = torch.as_tensor(start_qpos, dtype=torch.float32, device=self.device)
+        if qpos.dim() == 1:
+            qpos = qpos.unsqueeze(0)
+        return qpos.to(self.device).clone()
```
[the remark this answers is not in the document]
yangchen73: checked in the test

embodichain/lab/sim/planners/motion_generator.py
```diff
@@ -217,7 +220,11 @@ def generate(
...
 
-        options.plan_opts.control_part = options.control_part
+        if options.plan_opts is None:
+            if hasattr(self.planner, "default_plan_options"):
+                options.plan_opts = self.planner.default_plan_options()
+            else:
+                options.plan_opts = PlanOptions()
         result = self.planner.plan(
             target_states=target_plan_states, options=options.plan_opts
         )
```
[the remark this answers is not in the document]
yangchen73: Planner options go through plan_opts (same as ToppraPlanOptions)

docs/source/overview/sim/planners/neural_planner.md
```diff
@@ -0,0 +1,63 @@
+# NeuralPlanner
```
yuecideng, line 1: Mention that the policy of Neural Planner is specified for a robot.

yuecideng:
# Code Review: Neural Planner

> Automated review pass against commit `6168d0b` (current PR head). Focus: (1) the Neural Planner implementation, and (2) the structure of its integration into `MotionGenerator`.

## Verdict: **Merge with fixes**

No Critical issues — the code is functional, tests pass, and the planner-registry integration is structurally sound. Several **Important** issues remain, most clustered around the **MotionGenerator integration boundary**, which is the primary architectural question for this PR.

---

## Strengths

- **Clean planner-registry integration.** `embodichain/lab/sim/planners/motion_generator.py:96-99` registers `"neural": (NeuralPlanner, NeuralPlannerCfg)` via the same `_support_planner_dict` / `register_planner_type` pattern as `"toppra"`. `NeuralPlanner` subclasses `BasePlanner` exactly as `ToppraPlanner` does — the plug-in boundary at the planner level is correct and consistent with the established pattern.
- **Defensive checkpoint loading.** `embodichain/lab/sim/planners/neural_planner.py:223-291` validates every expected key (`agent`, `obs_normalizer.mean/var`), checks `policy_arch == "transformer"`, derives architecture dims from the checkpoint, and emits clear error messages listing available keys. More defensive than `ToppraPlanner`.
- **Correct quaternion conventions.** Traced the full path: `robot.compute_fk(to_matrix=False)` → `[pos, quat_wxyz]`; `_fk_pose_xyzw` converts wxyz→xyzw for the obs; `_is_active_reached` converts back to wxyz for `quat_error_magnitude`. All consistent.
- **Correct qpos-limits indexing** (`neural_planner.py:330-332`): `get_qpos_limits` returns `(N, dof, 2)`, correctly indexed to `(dof,2)` then `[:action_dim, 0/1]`.
- **Good asset-download UX** (`embodichain/data/assets/planner_assets.py:81-91`): actionable error with the three fix steps; proxy env-var normalization is a thoughtful touch.
- Tests run and pass (4 passed, CPU-only via `FakeRobot`). Conventions largely honored: copyright headers present on all new files; `from __future__ import annotations` + `__all__`; `@configclass` with `MISSING` for required `checkpoint_path`.

---

## Issues

### Important (should fix before merge)

**1. `is_interpolate=True` is incompatible with the neural planner and is unguarded**
`motion_generator.py:154-221` + `neural_planner.py:402-406`

`MotionGenerator.generate()`'s interpolation path converts EEF waypoints into `PlanState(move_type=JOINT_MOVE, qpos=...)` targets (lines 212-219) and passes them to `self.planner.plan()`. But `NeuralPlanner._parse_waypoints` rejects anything but `EEF_MOVE` with `xpos`. So a user who naively sets `is_interpolate=True` (a core `MotionGenerator` capability, used by every Toppra test) gets a `ValueError` from inside the planner rather than an upfront rejection.

*Why it matters:* This is the central integration finding. It shows the neural planner does not cleanly fit the "interpolate-then-plan" pipeline `MotionGenerator` is built around — TOPPRA consumes joint waypoints; the neural planner consumes raw EEF targets and runs its own rollout. The mismatch should be surfaced at the `MotionGenerator` boundary, not deep in `_parse_waypoints`.

*Fix:* Add a guard in `MotionGenerator.generate()` (or `__init__`) that rejects `is_interpolate=True` when the planner is `NeuralPlanner`, with a message explaining the neural planner runs its own waypoint rollout. Alternatively, document the contract on `MotionGenOptions.is_interpolate`.

**2. `control_part` / `start_qpos` contract change is a usability hazard**
`motion_generator.py:223-231`

The old line `options.plan_opts.control_part = options.control_part` was replaced by a comment saying planner-specific options "must be set on `plan_opts` explicitly." This means `MotionGenOptions.control_part` and `MotionGenOptions.start_qpos` — both real fields on `MotionGenOptions` (`motion_generator.py:55-59`) that Toppra users set routinely — are **silently ignored** on the neural path unless duplicated into `NeuralPlanOptions`. The example only sets them on `NeuralPlanOptions`, so a user porting the Toppra pattern will be surprised.

*Why it matters:* Two fields with the same name on two option objects, where only one is read depending on the planner, is a leaky abstraction. The removal of propagation is a behavior change with no deprecation path and no validation that `NeuralPlanOptions.control_part` is set until `plan()` runs (`neural_planner.py:319-324`).

*Fix:* Either propagate `MotionGenOptions.control_part`/`start_qpos` into `plan_opts` when the planner-specific options leave them `None`, or document the contract prominently and validate `control_part` at `MotionGenerator.generate()` entry rather than mid-rollout.

**3. `torch.load(..., weights_only=False)` is a security smell**
`neural_planner.py:229`

`weights_only=False` permits arbitrary code execution from the checkpoint. The checkpoint comes from a gated HuggingFace repo, which mitigates but does not eliminate the risk; torch ≥2.6 defaults to `weights_only=True` for this reason. The payload here is dicts/tensors/strings, all of which are safe under `weights_only=True`.

*Fix:* Try `weights_only=True` first; fall back to `False` only if a `UnpicklingError` occurs, with a logged warning.

**4. Public API docstrings missing — convention violation**
`neural_planner.py:205-206, 311-315`

`NeuralPlanner` has no class docstring and `plan()` has no docstring, while `ToppraPlanner` and `ToppraPlanner.plan()` both have Google-style docstrings. AGENTS.md requires "Google-style docstrings with Sphinx directives" on all public APIs. `_WaypointTransformerActor` and `_layer_init` also lack type annotations.

*Fix:* Add docstrings to `NeuralPlanner`, `plan`, `_load_checkpoint`. (`NeuralPlannerCfg`/`NeuralPlanOptions` field docs are good.)

**5. Leaky `PlanResult` contract: neural returns `velocities=None`/`accelerations=None`**
`neural_planner.py:379-385`

TOPPRA returns positions, velocities, accelerations, and a real `dt`/`duration` from the time-optimal solution. The neural planner returns only `positions` + `xpos_list`, fabricates a constant `dt` (cfg.dt=0.01) and `duration=(N-1)*dt`. `BasePlanner.is_satisfied_constraint` (`base_planner.py:140-191`) dereferences `vels.dtype` unconditionally — feeding it a neural result would crash. Nothing in the type system or docs signals this.

*Why it matters:* Consumers that treat `PlanResult` uniformly (the whole point of `MotionGenerator`'s abstraction) will break or silently get `None`. The fabricated `dt`/`duration` also misrepresent the trajectory's actual timing.

*Fix:* Either (a) document on `NeuralPlanner.plan`/the class that velocities/accelerations/dt are nominal-or-absent and guard `is_satisfied_constraint` against `None`, or (b) compute a defensible `dt` and leave velocities/accelerations `None` with a clear note.

**6. Tests verify plumbing, not planning behavior**
`tests/sim/planners/test_neural_planner.py:92-102, 146`

`FakeRobot.compute_fk(to_matrix=False)` returns a hardcoded `[[0,0,0,1,0,0,0]]` (origin + identity quat) regardless of `qpos`, and the test's waypoint is `torch.eye(4)` (also origin + identity). So `_is_active_reached` is trivially satisfied on step 0 — `success` is `True` no matter what actions the policy emits. `assert result.success is True` (line 146) proves the loop wiring, not that the policy drives the arm to a target. The shape/finiteness assertions (lines 147-151) are honest; the success assertion is not. Also, none of the three tests exercise the `MotionGenerator` integration failure modes (e.g. `is_interpolate=True` rejection, missing `control_part`).

*Fix:* Make `FakeRobot.compute_fk` depend on `qpos` (e.g. a trivial linear FK) so reaching a non-origin waypoint requires the policy to actually move, and add a test for the `is_interpolate=True` incompatibility once issue #1 is guarded.

### Minor (nice to have)

- **`num_arm_joints` cfg field is misleading** (`neural_planner.py:185`): the actor hard-raises unless `action_dim == 7` (`neural_planner.py:74-78`) and `_parse_obs` hardcodes 7-DoF slices throughout. The field is configurable but only `7` works. Either remove the field or gate it with a clear error.
- **Redundant quaternion round-trip** (`neural_planner.py:436-440, 488-489`): `_fk_pose_xyzw` builds xyzw, then `_is_active_reached` converts it back to wxyz. Keeping FK in wxyz and converting once for the obs would be cleaner.
- **Double FK per loop step** (`neural_planner.py:342, 362`): `compute_fk` called twice per iteration (obs + reach check). Cheap on CPU, but 2× solver calls on the GPU path.
- **`default_plan_options` asymmetry** (`neural_planner.py:220` vs Toppra): only `NeuralPlanner` implements it, so `MotionGenerator`'s `hasattr` fallback (`motion_generator.py:224`) is effectively neural-only. Adding it to `ToppraPlanner`/`BasePlanner` would make the `plan_opts=None` path uniform.
- **Test file missing `from __future__ import annotations`** (`test_neural_planner.py`): uses `str | None` annotations; works on 3.11 but violates the "every file" convention in AGENTS.md.
- **`_layer_init` return type annotation missing** (`neural_planner.py:42`): `def _layer_init(layer: nn.Linear, ...)` → add `-> nn.Linear`.

---

## Recommendations

1. **Decide explicitly whether `MotionGenerator` is the right host for a learned policy.** The planner registry is clean, but `NeuralPlanner` is semantically a *policy rollout*, not a *trajectory optimizer*. The `is_interpolate` incompatibility (#1) and the duplicated `control_part`/`start_qpos` (#2) are symptoms of forcing a policy into a pipeline built for optimizers. For an experimental feature, staying in `MotionGenerator` is acceptable *if* the incompatibilities are guarded and documented; if neural planning becomes permanent, consider a sibling `PolicyMotionGenerator` that bypasses the interpolation path entirely.
2. **Add an integration test** that drives `MotionGenerator` with a real (or non-trivial fake) Franka and asserts the end-effector actually converges to a non-trivial waypoint, not just that the loop runs.
3. **Run `black .`** before the next commit — the branch is at "WIP" and formatting hasn't been verified against `black==26.3.1`.

---

## Assessment

**Ready to merge: With fixes.**

The neural planner implementation itself is solid — checkpoint loading is defensive, quaternion/limit handling is correct, and tests pass. The integration into `MotionGenerator` is structurally sound at the planner-registry level (correct use of `BasePlanner`/`_support_planner_dict`), but it has real seams: the neural planner cannot consume the `JOINT_MOVE` targets that `MotionGenerator`'s interpolation path produces (unguarded, #1), the `control_part`/`start_qpos` contract was changed in a way that duplicates fields and silently ignores `MotionGenOptions` versions (#2), and the `PlanResult` is missing velocities/accelerations with no documented contract (#5). The tests verify plumbing rather than planning behavior (#6). Issues **#1, #2, #4, #6** should be resolved before merge; #3, #5 can land as follow-ups given the "experimental" labeling, but should not be dropped.

---
*Automated code review. Findings reviewed against commit `6168d0b`.*

embodichain/data/assets/planner_assets.py
```diff
@@ -0,0 +1,91 @@
+# ----------------------------------------------------------------------------
```
yuecideng, line 1: Rename to nmg_weights would be better. and place to a weights folder

embodichain/data/assets/__init__.py
```diff
@@ -21,3 +21,5 @@
 from .eef_assets import *
 from .robot_assets import *
 from .scene_assets import *
+from .solver_assets import *
```
yuecideng, line 24: Move solver as well

examples/sim/planners/neural_planner.py
```diff
@@ -0,0 +1,269 @@
...
+    )
+    target_states = [
+        PlanState(move_type=MoveType.EEF_MOVE, xpos=waypoint) for waypoint in waypoints
+    ]
+    result = motion_generator.generate(
+        target_states=target_states,
+        options=MotionGenOptions(
+            plan_opts=NeuralPlanOptions(
+                control_part=arm_name,
+                start_qpos=start_qpos,
```
yuecideng, line 243: How we handle start qpos now?

docs/source/overview/sim/planners/neural_planner.md
yangchen73: Mentioned

embodichain/data/assets/__init__.py
yangchen73: Moved

examples/sim/planners/neural_planner.py
yangchen73: Pass by `MotionGenOptions.start_qpos`. If not setted, use robot current qpos.

yuecideng (approved):

PR #341, opened 2026-06-29
Add Franka NMG benchmarks and demo path

skywhite1024:
## Description

This PR adds and wires up Franka neural motion generation benchmark coverage, including planner-level trials and the Franka pick-place demo/benchmark path.

It includes:
- Franka planner benchmarks for demo-matched offsets and reachable-FK waypoint banks.
- Franka pick-place demo/benchmark support with selectable IK/NMG planners and automated demo execution.
- Neural planner asset resolution and documentation updates for local/HuggingFace checkpoints.
- Atomic-action trajectory integration updates and tests for planner, trajectory, solver, and benchmark behavior.

Motivation: make it easier to compare IK-style planning against the experimental NMG backend on Franka, while keeping downstream pick-place validation separate from planner-only metrics.

Dependencies: adds `huggingface-hub` for neural planner checkpoint download support.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [x] Documentation update

## Screenshots

N/A

## Checklist

- [ ] I have run the `black .` command to format the code base.

### DIFF
--- a/docs/source/overview/sim/planners/index.rst
+++ b/docs/source/overview/sim/planners/index.rst
@@ -21,7 +21,7 @@ The `embodichain` project provides a unified interface for robot trajectory plan
 
 - **MotionGenerator**: A unified trajectory planning interface that supports joint/Cartesian interpolation, automatic constraint handling, flexible planner selection, and is easily extensible for collision checking and additional planners.
 - **ToppraPlanner**: A time-optimal trajectory planner based on the TOPPRA library, supporting joint trajectory generation under velocity and acceleration constraints.
-- **NeuralPlanner** (experimental): A learning-based EEF waypoint planner for Franka Panda.
+- **NeuralPlanner**: An experimental neural motion generation backend for end-effector waypoint trajectories, backed by a trained waypoint-conditioned policy.
 - **TrajectorySampleMethod**: An enumeration for trajectory sampling strategies, supporting sampling by time, quantity, or distance.
 
 These tools can be used to generate smooth and dynamically feasible robot trajectories, and are extensible for future collision checking and various sampling requirements.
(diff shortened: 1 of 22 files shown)


skywhite1024:
After switching the demo to the Franka arm + DH_PGI gripper, `ik_interpolate` can plan and physically lift the object, but the NMG planner still fails before replay. The observed failure is at the arm approach planning stage:
NeuralPlanner failed to plan arm trajectory for env 0.
PickUp failed to plan the approach trajectory.
This suggests the issue is not the gripper contact model anymore. The failure happens before physical grasp execution.
My current hypothesis is that this is mostly an out-of-distribution generalization issue for the current `franka.pt` NMG checkpoint. `ik_interpolate` solves each EEF waypoint geometrically against the live robot model/TCP, so it naturally adapts to the DH hybrid gripper and the selected grasp pose. NMG, however, is a learned waypoint policy. The hybrid DH gripper changes the effective TCP/tool geometry and produces grasp/approach poses that differ from the training distribution. Atomic pick-place also creates task-specific waypoints from grasp affordances, which are stricter than the planner-only FK/demo-offset benchmark.
Suggested improvements:
1. Add clearer NMG failure diagnostics: checkpoint path, start qpos, target waypoint pose, terminal pose error, active waypoint index, and whether rollout hit `max_steps`.
2. Try `neural_refine` as the default neural demo mode, or add IK snapping/projection for final key waypoints.
3. Rerank top-K grasp candidates by NMG reachability instead of selecting only the best antipodal/IK candidate.
4. Add fallback behavior: use NMG for free-space motion, but use IK for grasp approach/close/lift when NMG fails.
5. Fine-tune or retrain NMG on the Franka + DH gripper TCP distribution and atomic pick-place waypoints.
6. Include tool/TCP extrinsics in the NMG observation, or randomize TCP offsets during training to improve robustness to gripper changes.
7. Distill successful IK/TOPPRA pick-place trajectories into NMG training data.
8. Evaluate NMG separately on planner-only reachability and atomic pick-place, since the latter is much more sensitive to small TCP/pose errors.
[Uploading Kooha-2026-06-30-11-22-25.webm…]()

PR #436, opened 2026-07-29
Add scene engine  pipeline

MuziWong:
## Description

  This PR adds the image to tabletop scene generation pipeline, including semantic understanding, image segmentation and verification, geometry generation, layout refinement, gravity settling, Gym export, and preview tooling.

  Dependencies: Existing scene-engine service endpoints and simulation dependencies.

  ## Type of change

  - [ ] Bug fix
  - [ ] Enhancement
  - [x] New feature
  - [ ] Breaking change
  - [ ] Documentation update

  ## Screenshots

  N/A

  ## Checklist

  - [x] I have run the `black .` command to format the code base. (`black` is unavailable in the current
  environment; the branch includes its existing `RAN black` commit.)
  - [x] I have added tests that prove my fix is effective


embodichain/gen_sim/scene_engine/cli/start.py
```diff
@@ -0,0 +1,70 @@
...
+    resolved_output_root.mkdir(parents=True, exist_ok=True)
+
+    generate_scene_from_image(
+        image_path=resolved_image_path,
+        output_root=resolved_output_root,
+    )
+    print("Successfully completed!")
+
+
+def main() -> None:
```
yuecideng: [P2] 请将新增命令注册到统一 CLI。当前项目公开入口是 `embodichain` console script，但 `embodichain.__main__.COMMANDS` 中没有 Scene Engine 或预览命令；因此安装后用户无法从统一 CLI 发现或调用这里的 `main()`，只能依赖内部模块路径。请在 `COMMANDS` 中注册 Scene Engine 和 preview 对应的子命令。

embodichain/gen_sim/scene_engine/cli/start.py
```diff
@@ -0,0 +1,70 @@
...
+        raise ValueError(f"Image input is not a file: {resolved_image_path}")
+    if resolved_image_path.suffix.lower() not in _SUPPORTED_IMAGE_SUFFIXES:
+        raise ValueError(
+            "Image input must have one of these extensions: .jpg, .jpeg, .png"
+        )
+
+    resolved_output_root = Path(output_root).expanduser().resolve()
+    resolved_output_root.mkdir(parents=True, exist_ok=True)
+
+    generate_scene_from_image(
```
yuecideng, line 46: [P1] Pass service configuration through the generation CLI. As written, users cannot run this command without editing the JSON installed with the package: this call forwards only the image and output paths, while the packaged service URLs are empty. The LLM endpoint can be overridden through environment variables, but the segmentation and geometry services have no CLI or environment-variable entry point, so generation fails during `base_url` validation. Please accept a config path or explicit service URL options here and forward them into the pipeline.

embodichain/gen_sim/scene_engine/pipeline/gym_export.py
```diff
@@ -0,0 +1,220 @@
...
+
+    exported_entries = {
+        scene_object.id: _copy_scene_object_to_gym_assets(
+            scene_object=scene_object,
+            mesh_assets_root=mesh_assets_root,
+        )
+        for scene_object in scene_objects
+    }
+    gym_config = {
+        "id": f"Prompt2Scene-{int(time.time() * 1000)}-v0",
```
yuecideng: [P1] Export a registered environment ID. `embodichain run-env` passes this generated ID directly to `gymnasium.make`, but the codebase does not register any `Prompt2Scene-*` environment. Consequently, every exported config fails immediately with `NameNotFound`. Please export an environment ID that is already registered with the runner, or provide and invoke the corresponding registration logic.

embodichain/gen_sim/scene_engine/pipeline/gym_export.py
```diff
@@ -0,0 +1,220 @@
...
+        "max_episodes": 10,
+        "max_episode_steps": 300,
+        "env": {"events": {}, "observations": {}, "dataset": {}},
+        "robot": {},
```
yuecideng: [P1] Supply a valid robot configuration in the Gym export. Even after the environment ID is fixed, `EmbodiedEnv` parses this empty object as `RobotCfg(fpath=None)`; `_setup_robot` then receives `None` from `add_robot` and proceeds to access robot joints. If this file is intended to load directly through `run-env`, export a valid robot configuration. Otherwise, target an explicitly scene-only environment that supports running without a robot.

embodichain/gen_sim/scene_engine/pipeline/scene_export.py
```diff
@@ -0,0 +1,220 @@
...
+
+    source_glb_path = Path(scene_object.simready_glb_path).expanduser().resolve()
+    if not source_glb_path.is_file():
+        raise FileNotFoundError(
+            f"SimReady GLB for scene object {object_id!r} not found: {source_glb_path}"
+        )
+    destination_glb_path = mesh_assets_root / object_id / f"{object_id}.glb"
+    destination_glb_path.parent.mkdir(parents=True, exist_ok=True)
+    shutil.copy2(source_glb_path, destination_glb_path)
+    return destination_glb_path.relative_to(mesh_assets_root.parent).as_posix()
```
yuecideng: [P1] Emit mesh paths that the Gym loader can resolve. When this config is loaded through `run-env`, `config_to_cfg` passes the relative Mesh path to `get_data_path`, which interprets `mesh_assets` as a dataset name and raises `Dataset class 'mesh_assets' not found`. Only the custom preview resolves this path relative to the config directory. Please export an absolute or data-root-relative path, or update the common loader to resolve Mesh paths relative to the config file.

MuziWong: Resolved by changing this artifact from a run-env Gym configuration to an explicit scene export (`embodichain.scene-export/v1`). The output is now written as `scene_export/scene_config.json` and is loaded through the Scene Engine preview path, which resolves `mesh_assets/...` relative to the config file directory.

It is no longer passed to `config_to_cfg()` or `get_data_path()`, so exporting absolute paths or modifying the common Gym loader is unnecessary here. Keeping config-relative paths also preserves portability when the complete `scene_export/` directory is moved.

embodichain/gen_sim/scene_engine/pipeline/gym_export.py
MuziWong: Resolved by changing the artifact from a Gym environment configuration to an explicit scene export (`embodichain.scene-export/v1`).

The generated scene does not infer a robot model, robot placement, controller, or sensor setup from the input image, so exporting a default robot configuration would produce an incomplete and potentially misleading task setup. The scene export therefore omits the `robot` field entirely and is loaded through the Scene Engine preview path, which directly creates a `SimulationManager` for the table and assets without constructing an `EmbodiedEnv`.

A downstream task or environment layer can attach an explicit robot configuration when one is available, keeping scene generation independent from task-specific execution.

embodichain/gen_sim/scene_engine/pipeline/gym_export.py
MuziWong: Resolved by changing the artifact from a Gym environment configuration to an explicit scene export (`embodichain.scene-export/v1`).

The generated identifier is now `scene_id`, which uniquely identifies the exported scene data. It is not a Gymnasium environment ID and is never passed to `gymnasium.make()`. The exported scene is loaded through the Scene Engine preview path instead of `embodichain run-env`, so no Gym environment registration is required.

embodichain/gen_sim/scene_engine/cli/start.py
MuziWong: Resolved by adding an optional `--config` argument to the generation CLI. The provided Scene Engine JSON configuration is propagated to the LLM, image segmentation, and geometry generation clients, allowing users to override service endpoints without modifying the package-installed default configuration.

embodichain/gen_sim/scene_engine/cli/start.py
MuziWong: Resolved by registering `scene-engine` and `preview-scene` in `embodichain.__main__.COMMANDS`.

The unified dispatcher lazily imports the selected command handler and forwards the remaining command-line arguments to its `main(argv)` function. Both Scene Engine CLI entry points accept the forwarded `argv`, allowing their own argument parsers to process the expected options instead of relying on the process-wide `sys.argv`.

Installed users can invoke Scene Engine through the public CLI:

```bash
embodichain scene-engine \
  --image <image> \
  --output_root <output_root> \
  --config <scene_engine_config>

embodichain preview-scene \
  --output_root <output_root>
```

`preview-scene` opens the native preview window by default. To use the browser-based Viser preview:

```bash
embodichain preview-scene \
  --output_root <output_root> \
  --viser \
  --viser-host 0.0.0.0 \
  --viser-port 9000
```

The corresponding module entry points are:

```bash
python -m embodichain.gen_sim.scene_engine.cli.start \
  --image <image> \
  --output_root <output_root> \
  --config <scene_engine_config>

python -m embodichain.gen_sim.scene_engine.cli.preview \
  --output_root <output_root>
```

I also verified that:

- `scene-engine` and `preview-scene` appear in `embodichain --help`;
- `embodichain scene-engine --help` and `embodichain preview-scene --help` dispatch to their command-specific parsers correctly;
- `preview-scene --help` exposes the required `--output_root` argument and the optional Viser-related flags;
- the relevant CLI and Scene Engine unit tests pass (`21 passed`).

embodichain/gen_sim/scene_engine/pipeline/scene_understanding.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+def validate_scene_understanding(scene: Scene) -> None:
+    """Validate that scene understanding produced a complete semantic scene."""
+    if scene.table is None:
+        raise ValueError("Scene understanding must identify a table.")
+    if (
+        scene.table.id != "table"
+    ):  # Currently it will always return true. For we hardcode the table id to "table".
+        raise ValueError("Scene table id must be 'table'.")
```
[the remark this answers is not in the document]
MuziWong: I have already shorten this into one line.

embodichain/gen_sim/scene_engine/pipeline/utils/scene_segmentation_utils.py
```diff
@@ -0,0 +1,340 @@
...
+        mask = decode_rle_mask(candidate.mask_rle)
+        _require_image_size(mask, image.size)
+        decoded_masks.append((candidate, mask))
+        color_layer = Image.new(
+            "RGBA", image.size, colors[(candidate.index - 1) % len(colors)]
+        )
+        transparent_layer = Image.new("RGBA", image.size, (0, 0, 0, 0))
+        rendered_mask = (  # If weuse outline, then need to do some another processings.
+            mask if mask_style == "fill" else _mask_outer_outline(mask, image.size)
+        )
```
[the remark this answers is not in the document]
MuziWong: I have changed weuse into we use.

embodichain/gen_sim/scene_engine/clients/geometry_generation.py
```diff
@@ -0,0 +1,374 @@
...
+        for (
+            object_id,
+            _,
+        ), response_object in zip(  # Pair each object_id with its response_object for downloading the glb.
+            resolved_object_masks,
+            response_objects,
+        ):
+            output_path = resolved_output_root / f"{object_id}.glb"
+            self._download_glb(response_object["mesh"], output_path)
+        return response_data, response_objects
```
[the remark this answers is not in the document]
MuziWong: I have already fixed this.

embodichain/gen_sim/scene_engine/cli/preview.py
```diff
@@ -0,0 +1,202 @@
...
+        print(f"Previewing: {config_path}")
+        print("Close with Ctrl-C.")
+        sim.open_window()
+        while True:
+            time.sleep(0.1)
+    except KeyboardInterrupt:
+        print("Stopping preview.")
+    finally:
+        sim.destroy()
+
```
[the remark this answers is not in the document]
MuziWong: I have already fixed this part.

embodichain/gen_sim/scene_engine/pipeline/generate.py
```diff
@@ -0,0 +1,118 @@
...
+    image_segmentation_client.check_health()  # Error raising will happen internally.
+    scene = segment_scene(
+        image_path=image_path,
+        output_root=resolved_output_root,
+        scene=scene,
+        vlm_client=vlm_client,
+        image_segmentation_client=image_segmentation_client,
+    )
+    image_segmentation_client.close()  # Kill the session.
+    log_stage_end("Scene Segmentation")
```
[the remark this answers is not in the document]
MuziWong: I have already fixed this part.

embodichain/gen_sim/scene_engine/pipeline/utils/scene_generation_utils.py
```diff
@@ -0,0 +1,1549 @@
...
+            )
+            scale_matrix = np.eye(4)
+            scale_matrix[:3, :3] = np.diag(prepared_assets[asset_id]["z_up_scale"])
+            final_z_up_layout_matrix = final_rigid_pose_z_up @ scale_matrix
+            settled_layout_by_id[asset_id] = transform_matrix_to_layout_object(
+                asset_id,
+                z_up_to_y_up_matrix @ final_z_up_layout_matrix @ y_up_to_z_up_matrix,
+            )
+    finally:
+        sim._deferred_destroy()
```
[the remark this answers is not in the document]
MuziWong: I have already fixed this part.

embodichain/gen_sim/scene_engine/pipeline/generate.py
```diff
@@ -0,0 +1,118 @@
...
+
+def generate_scene_from_image(
+    image_path: str | Path,
+    output_root: str | Path,
+    *,
+    llm_config_path: str | Path | None = None,
+    image_segmentation_config_path: str | Path | None = None,
+    geometry_generation_config_path: str | Path | None = None,
+) -> Scene:
+    """Generate the initial core scene state from an input image."""
```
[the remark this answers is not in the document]
MuziWong: Thank you for reviewing!

embodichain/gen_sim/scene_engine/cli/preview.py
```diff
@@ -0,0 +1,212 @@
...
+        print(f"Previewing: {config_path}")
+        print("Close with Ctrl-C.")
+        sim.open_window()
+        while True:
+            time.sleep(0.1)
+    except KeyboardInterrupt:
+        print("Stopping preview.")
+    finally:
+        sim.destroy(exit_process=False)
+        _EmbodiSimManager.flush_cleanup_queue()
```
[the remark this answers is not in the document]
MuziWong: Thanks your review, I have already fix that part.

embodichain/gen_sim/scene_engine/utils/logger.py
```diff
@@ -0,0 +1,38 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ----------------------------------------------------------------------------
+
+
+from __future__ import annotations
```
yuecideng, line 19: Can we use embodichain logger?

MuziWong: Thank you for your review. We have already fixed this problem. Now we use the offical logger of EmbodiChain.

embodichain/gen_sim/scene_engine/cli/start.py
```diff
@@ -0,0 +1,93 @@
...
+    )
+    parser.add_argument(
+        "--config",
+        type=Path,
+        default=None,
+        help=(
+            "Optional Scene Engine JSON config containing the llm, "
+            "image_segmentation, and geometry_generation service settings."
+        ),
+    )
```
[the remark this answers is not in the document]
MuziWong: Thank you for your reviewing, I have already fix this problem and updated relevant docs :)

embodichain/gen_sim/scene_engine/cli/preview.py
```diff
@@ -0,0 +1,236 @@
...
+        if not isinstance(shape, dict) or not isinstance(shape.get("fpath"), str):
+            raise ValueError(f"Scene {label} {uid!r} has no shape.fpath.")
+        if shape.get("shape_type") != "Mesh":
+            raise ValueError(
+                f"Scene {label} {uid!r} must use shape_type='Mesh' for preview."
+            )
+
+        mesh_path = (config_dir / shape["fpath"]).resolve()
+        if not mesh_path.is_file():
+            raise FileNotFoundError(f"Gym mesh for {uid!r} not found: {mesh_path}")
```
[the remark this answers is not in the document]
MuziWong: Thank you for your review. I have fix this problem and updated the test script.

embodichain/gen_sim/scene_engine/clients/geometry_generation.py
```diff
@@ -0,0 +1,476 @@
...
+        # , which is safe because we validated the lengths earlier.
+        for (
+            object_id,
+            _,
+        ), response_object in zip(  # Pair each object_id with its response_object for downloading the glb.
+            resolved_object_masks,
+            response_objects,
+        ):
+            output_path = resolved_output_root / f"{object_id}.glb"
+            self._download_glb(response_object["mesh"], output_path)
```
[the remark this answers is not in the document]
MuziWong: Thank you, dear Copilot~ I have fix this problem and updated the test script.

docs/source/features/generative_sim/scene_engine.md
```diff
@@ -0,0 +1,156 @@
+# Scene Engine
+
+The Scene Engine converts one tabletop-scene image into a scene-only export. It
+identifies a table and visible assets, generates their meshes, refines their
+layout, settles them under gravity, and writes an EmbodiChain scene export.
+
+## Quick Start
+
+Install EmbodiChain with the generative-simulation dependencies. See
+[Installation (gensim extra)](../../quick_start/install.md#optional-generative-simulation-gensim).
```
yuecideng: [P1] The `gensim` extra currently declares only `bpy` and `pyrender`, while Scene Engine directly imports `requests`, Pillow, SciPy, trimesh, Open3D, and matplotlib. A clean `embodichain[gensim]` installation therefore relies on undeclared transitive dependencies. Please declare these dependencies directly, possibly under a dedicated `scene-engine` extra, and exercise that installation path in CI.

embodichain/gen_sim/scene_engine/core/asset.py
```diff
@@ -0,0 +1,51 @@
...
+    category: str
+    name: str
+    description: str
+    # Path to a binary mask image aligned with the input image. White pixels
+    # identify this asset; black pixels identify the background.
+    mask_path: str | None = None
+    # Absolute path to the canonicalized GLB used by the final simulation.
+    simready_glb_path: str | None = None
+    # Final y-up layout after scene refinement and gravity settling.
+    rot: list[float] | None = None
```
yuecideng: [P1] This mutable dataclass represents every pipeline stage through nullable fields, while `Table` duplicates almost the same shape. That makes invalid state combinations representable until a late export-time check. Please consider a shared `SceneObject` model with typed pose/frame/artifact fields, or stage-specific validated models. The scene IR will also need support relations and confidence/provenance if stacked or contained objects are to be preserved.

embodichain/gen_sim/scene_engine/pipeline/generate.py
```diff
@@ -0,0 +1,121 @@
...
+from embodichain.gen_sim.scene_engine.pipeline.scene_export import export_scene
+
+
+def generate_scene_from_image(
+    image_path: str | Path,
+    output_root: str | Path,
+    *,
+    llm_config_path: str | Path | None = None,
+    image_segmentation_config_path: str | Path | None = None,
+    geometry_generation_config_path: str | Path | None = None,
```
yuecideng: [P1] One Scene Engine JSON file is exposed here as three independently parsed paths, while important behavior such as IoU thresholds, retry counts, packing clearances, settle steps, physics settings, and hull counts remains hard-coded elsewhere. Please load and validate one project-style `@configclass` `SceneEngineCfg`, then inject the relevant sub-config into each stage/client. This would also make runs reproducible through a single config snapshot.

embodichain/gen_sim/scene_engine/pipeline/scene_generation.py
```diff
@@ -0,0 +1,574 @@
...
+    *,
+    vlm_client: OpenAICompatibleVLM,
+    geometry_generation_client: GeometryGenerationClient,
+) -> Scene:
+
+    resolved_image_path = _validate_image_path(image_path)
+    # Create stage output directory.
+    stage_output_root = Path(output_root).expanduser().resolve() / "scene_generation"
+    if stage_output_root.exists():
+        shutil.rmtree(stage_output_root)
```
yuecideng, line 73: [P1] This removes the previous, potentially expensive generation artifacts before a replacement has completed. Any service or simulation failure leaves a partial stage and prevents resuming. Please build each stage in a temporary directory, validate it, write an input/config/model manifest, and atomically replace the completed stage. A `--resume`/`--from-stage` path would be especially valuable for this multi-service pipeline.

embodichain/gen_sim/scene_engine/pipeline/utils/scene_generation_utils.py
```diff
@@ -0,0 +1,1542 @@
...
+        projected_rectangles_by_id[asset_id] = asset_2d_aabb.tolist()
+
+    # 2. Project every table triangle into the z-up world's x-y plane.
+    if len(table_world_mesh.vertices) < 3 or len(table_world_mesh.faces) == 0:
+        raise ValueError("Table geometry must contain at least one triangle.")
+    projected_vertices = table_world_mesh.vertices[
+        :, :2
+    ]  # Ignore z, for we wanna get the x-y plane projection.
+    try:
+        projected_hull = ConvexHull(
```
yuecideng: [P0] This is the convex hull of the projection of the entire table mesh, not the actual top support surface. It fills concavities and holes and may include footprints from legs or overhanging geometry, so the packer can place assets over unsupported space. Please extract near-top, upward-facing triangles (or a support height field/polygon) and validate candidate footprints with downward raycasts. The existing comment about L-shaped tables indicates this is a known correctness limitation.

embodichain/gen_sim/scene_engine/pipeline/utils/scene_generation_utils.py
```diff
@@ -0,0 +1,1542 @@
...
+    )
+    boundary_size = boundary_half_extents * 2.0
+    if np.any(boundary_size <= 0):
+        raise ValueError(
+            "Table support-region boundary must have non-zero width and height."
+        )
+
+    # Keep the internal rectangle centered at the table/world origin.
+    # rectangle_center = convex_boundary.mean(axis=0) # The mean is not always 0,0.
+    rectangle_center = np.array([0.0, 0.0])
```
yuecideng: [P1] Despite the function name, this is not the largest internal axis-aligned rectangle: it is constrained to the world origin, a fixed AABB-derived aspect ratio, and world XY alignment. Rotated or asymmetric tables can lose substantial usable area or reject an otherwise feasible layout. Please either rename/document this precise heuristic or optimize directly against the support polygon, optionally using an oriented rectangle.

embodichain/gen_sim/scene_engine/pipeline/utils/scene_generation_utils.py
```diff
@@ -0,0 +1,1542 @@
...
+            y_up_to_z_up_matrix=y_up_to_z_up_matrix,
+        )
+        asset_world_mesh = _mesh_to_z_up_world_for_aabb(
+            y_up_mesh=asset_mesh,
+            z_up_rigid_layout=asset_rigid_layout,
+            z_up_scale=asset_z_up_scale,
+            y_up_to_z_up_matrix=y_up_to_z_up_matrix,
+        )
+        asset_bottom_z = float(asset_world_mesh.bounds[0, 2])
+        asset_rigid_layout["pos"][2] += table_top_z + clearance - asset_bottom_z
```
yuecideng: [P0] Moving every asset bottom independently to the table top destroys vertical relationships that the earlier group alignment explicitly preserves. Combined with planar AABB separation, a mug on a book, an item in a tray, or any stacked pair becomes a set of unrelated table-level objects. Please infer/store a support graph and settle objects per support parent/layer instead of flattening every asset onto the table.

embodichain/gen_sim/scene_engine/pipeline/utils/scene_generation_utils.py
```diff
@@ -0,0 +1,1542 @@
...
+                allowed_max=allowed_max,
+                aabb_clearance=aabb_clearance,
+                tolerance=tolerance,
+            )
+            if candidates is None:
+                continue
+            if not candidates:
+                # Both AABBs are already blocked by the table boundary on every
+                # separating axis. Keep the current boundary-safe layout and
+                # let the later gravity simulation handle this residual overlap.
```
yuecideng: [P0] Returning an overlapping layout silently delegates an invalid initial state to the physics solver. Deep interpenetration can produce explosive depenetration, toppled assets, or objects leaving the table; gravity is not a reliable packing solver. Please treat unresolved overlap as an explicit retry/failure, or use a constrained packing solver that can report infeasibility.

embodichain/gen_sim/scene_engine/pipeline/utils/scene_generation_utils.py
```diff
@@ -0,0 +1,1542 @@
...
+                    body_scale=tuple(asset_info["y_up_scale"]),
+                    body_type="dynamic",
+                    max_convex_hull_num=max_convex_hull_num,
+                    acd_method="vhacd",  # Use vhacd by default.
+                )
+            )
+
+        # All assets share this one simulation, so they can collide with the
+        # table and with one another while settling.
+        sim.update(step=settle_steps)
```
yuecideng: [P0] A fixed number of simulation steps is not a convergence or validity criterion. Before exporting, please require linear/angular velocities to remain below thresholds for several frames and validate finite poses, table support, footprint containment, and unacceptable penetration/fall-off. If those postconditions fail, the pipeline should retry layout/settling or return a clear quality failure rather than exporting the pose.

embodichain/gen_sim/scene_engine/pipeline/scene_export.py
```diff
@@ -0,0 +1,220 @@
...
+from embodichain.gen_sim.scene_engine.core.table import Table
+
+_DEFAULT_MAX_CONVEX_HULL_NUM = 16
+_TABLE_PHYSICS_ATTRS = {
+    "mass": 10.0,
+    "static_friction": 0.95,
+    "dynamic_friction": 0.9,
+    "restitution": 0.01,
+}
+_ASSET_PHYSICS_ATTRS = {
```
yuecideng: [P1] The settling simulation uses `RigidObjectCfg` defaults and 32 convex hulls, while the exported assets use these different physical attributes and 16 hulls. The final dynamic scene may therefore not reproduce the behavior under which its pose was settled. Please share the same object physics and collision-decomposition configuration between settlement, export, and downstream loading.

embodichain/gen_sim/scene_engine/pipeline/scene_export.py
```diff
@@ -0,0 +1,220 @@
...
+        field_name="table_max_convex_hull_num",
+    )
+    asset_max_convex_hull_num = _positive_int(
+        asset_max_convex_hull_num,
+        field_name="asset_max_convex_hull_num",
+    )
+
+    export_root = Path(output_root).expanduser().resolve() / "scene_export"
+    mesh_assets_root = export_root / "mesh_assets"
+    mesh_assets_root.mkdir(parents=True, exist_ok=True)
```
yuecideng: [P1] `scene_export` is neither refreshed nor written atomically. Re-running into the same output root with fewer objects leaves stale mesh directories, and a copy/write failure can leave a config and asset set from different runs. Please assemble a fresh temporary export, validate all entries, then atomically replace the previous directory.

embodichain/gen_sim/scene_engine/clients/geometry_generation.py
```diff
@@ -0,0 +1,486 @@
...
+        return response_data, response_objects
+
+    def _request_objects(
+        self,
+        *,
+        image_path: Path,
+        object_masks: list[tuple[str, Path]],
+    ) -> tuple[dict[str, Any], list[dict[str, Any]]]:
+        last_error: Exception | None = None
+        for _ in range(self._max_attempts):
```
yuecideng, line 163: [P1] Retrying the entire submission immediately can create duplicate expensive geometry jobs when the server accepted the request but the response was lost. Please separate submit/poll/download retry policies, add exponential backoff and jitter, and use an idempotency key or persisted request ID. The polling endpoint, interval, and 600-second deadline should also come from configuration.

tests/gen_sim/scene_engine/test_scene_generation_utils.py
```diff
@@ -0,0 +1,102 @@
...
+            [minimum[0], minimum[1]],
+            [maximum[0], minimum[1]],
+            [maximum[0], maximum[1]],
+            [minimum[0], maximum[1]],
+        ],
+        dtype=float,
+    )
+
+
+def test_layout_transform_round_trip_preserves_pose_and_scale() -> None:
```
yuecideng: [P1] The current tests cover a transform round trip and a simple AABB happy path, but not the high-risk behavior introduced here. Please add fixtures for concave/rotated/asymmetric tables, stacked or contained assets, dense/infeasible packing, gravity convergence and fall-off checks, queued geometry jobs, and a mocked end-to-end pipeline. Golden-scene metrics such as support validity and relation retention would make scene quality regressions visible.

yuecideng (commented):
The overall layering is strong: the staged pipeline, strict service-boundary validation, explicit coordinate-frame handling, debug artifacts, and portable scene export form a solid end-to-end MVP. The inline comments below focus on the gaps that should be addressed before treating this as a robust scene reconstruction engine. P0 marks scene-correctness risks; P1 marks reliability and maintainability risks.

XuanchaoPENG:
Next you should think about how to add the articulation into the scene engine. A possible solution is to add a special signal to use the articraft-style articulation pipeline. Then put it into the scene by using 2d optimization. Whether generating articulation depends on whether we want to interact with this object.

pyproject.toml
```diff
@@ -52,9 +52,27 @@ dependencies = [
...
+  "pyrender==0.1.45",
+  "requests",
+  "Pillow",
+  "numpy",
+  "scipy",
+  "shapely",
+  "trimesh",
+  "open3d",
+  "matplotlib"
 ]
```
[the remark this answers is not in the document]
MuziWong: Thanks — addressed.

Scene Engine and SimReady Pipeline are peer components of the `gen_sim` module, so we kept `gensim` as their shared installation extra rather than adding a separate `scene-engine` extra.

### Changes
- Kept `gensim` as the shared installation extra for both Scene Engine and SimReady Pipeline.
- Explicitly added Scene Engine's direct runtime dependencies to the `gensim` extra:
  - `requests`
  - `Pillow`
  - `numpy`
  - `scipy`
  - `shapely`
  - `trimesh`
  - `open3d`
  - `matplotlib`
- Fixed the missing TOML commas.
- Added a CI import/install check for the `gensim` installation path.

embodichain/gen_sim/scene_engine/pipeline/utils/scene_exporter.py
```diff
@@ -0,0 +1,195 @@
...
+    def _copy_scene_object_to_assets(
+        *,
+        scene_object: SceneObject,
+        mesh_assets_root: Path,
+    ) -> str:
+        """Copy one referenced SimReady GLB and return its config-relative path."""
+        object_id = scene_object.id
+        if Path(object_id).name != object_id or object_id in {"", ".", ".."}:
+            raise ValueError(
+                f"Scene object id is not safe for a GLB filename: {object_id!r}"
```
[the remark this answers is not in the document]
MuziWong: - Addressed. `SceneExporter` now rejects backslashes in `object_id`, matching `GeometryGenerationClient`'s filename-safety checks. This prevents Windows path separators from bypassing validation and escaping the intended output directory.
- Added a regression test for malicious IDs such as `..\\evil`.

embodichain/gen_sim/scene_engine/llms/openai_compatible_client.py
```diff
@@ -0,0 +1,139 @@
...
+from typing import Any
+from urllib.error import HTTPError, URLError
+from urllib.request import Request, urlopen
+
+from embodichain.gen_sim.scene_engine.llms.load_config import LLMConfig, load_llm_config
+
+_SUPPORTED_IMAGE_SUFFIXES = {".jpg", ".jpeg", ".png"}
+
+
+class OpenAICompatibleVLM:
```
yuecideng, line 32: Could we build a unified MLLM interface for all use case across EmbodiChain

pyproject.toml
```diff
@@ -54,7 +54,15 @@ dependencies = [
 [project.optional-dependencies]
 gensim = [
   "bpy",
-  "pyrender==0.1.45"
+  "pyrender==0.1.45",
+  "requests",
+  "Pillow",
+  "numpy",
```
yuecideng: numpy, shapely, trimesh and open3d already included in dexsim. So please remove them

embodichain/gen_sim/scene_engine/llms/openai_compatible_client.py
MuziWong: Yes, I agree that we could eventually provide a unified entry point, either under the `gen_sim` module or at the EmbodiChain level.

However, the Scene Engine, Action Engine, and Asset Engine are still under active development. For now, I suggest keeping their entry points and registrations separate, and then unifying them in a follow-up PR once their interfaces and responsibilities are more stable.

What do you think?

pyproject.toml
MuziWong: Updated `pyproject.toml` as requested.

### Changes
- Removed the following dependencies from the `gensim` extra because they are already provided by `dexsim_engine`:
  - `numpy`
  - `shapely`
  - `trimesh`
  - `open3d`
- Kept the remaining Scene Engine dependencies explicitly declared in the `gensim` extra.
- Verified that the optional dependency list remains valid TOML.

yuecideng (approved):

PR #438, opened 2026-07-29
feat(w1): add version-aware V025 support

yhnsu:
# Description

Add version-aware Dexforce W1 support for V021, V022, and V025.

Key changes:

- Add `W1VersionSpec` as the source of truth for versioned assets, kinematics, sensors, EEF attachment, and TCP configuration.
- Register unified V022 and V025 asset archives for automatic download from Hugging Face.
- Add the V025 12 mm version-owned EEF offset and apply it consistently to URDF assembly and FK/IK TCP.
- Support default and custom end effectors without duplicating version corrections.
- Remove the industrial-arm variant and `arm_kind`; W1 now uses the anthropomorphic arm model directly.
- Update the W1 examples, documentation, and configuration tests.

Asset paths:

```text
dexforce_w1/v022/w1.zip
dexforce_w1/v025/w1.zip
```

No new Python dependencies are required.

## Type of change

- Breaking change (industrial-arm and `arm_kind` configurations are removed)
- Documentation update

## Screenshots

https://github.com/user-attachments/assets/4f374015-fae6-4b6a-8292-ec596d922838

## Testing

```text
tests/sim/objects/test_robot_cfg.py: 34 passed
```

## Checklist

- [x] No dependency updates are required.


embodichain/lab/sim/robots/dexforce_w1/specs.py
```diff
@@ -0,0 +1,269 @@
...
+def normalize_component_versions(
+    versions: dict[DexforceW1Type | str, DexforceW1Version | str] | None,
+) -> dict[DexforceW1Type, DexforceW1Version]:
+    normalized = {}
+    for component_type, version in (versions or {}).items():
+        if not isinstance(component_type, DexforceW1Type):
+            component_type = DexforceW1Type(component_type)
+        if not isinstance(version, DexforceW1Version):
+            version = DexforceW1Version(version.lower())
+        normalized[component_type] = version
```
yuecideng: **Blocking: `component_versions` currently cannot round-trip.** The key is normalized to `DexforceW1Type` here, but `RobotCfg.to_dict()` serializes dict keys with `str(k)`, so `{"left_arm": "v025"}` becomes `{"DexforceW1Type.LEFT_ARM": "v025"}`; on the next `from_dict()`, the `DexforceW1Type(component_type)` call on the line above raises `ValueError`. I recommend explicitly using the enum's `.value` when serializing (or making the base serializer handle enum keys correctly), and adding a round-trip test that includes `component_versions`.

embodichain/lab/sim/robots/dexforce_w1/utils.py
```diff
@@ -733,11 +684,19 @@ def build_dexforce_w1_cfg(
     from embodichain.lab.sim.robots.dexforce_w1.cfg import DexforceW1Cfg
 
     cfg = DexforceW1Cfg()
```
yuecideng: **Blocking: this public builder does not encode the build options back into the config object.** For example, with `include_hand=False`, the returned object's `with_default_eef` is still `True`; `arm_sides`, `include_head/eyes/wrist_cameras`, and the hand type/version have no corresponding fields either. In practice, after `build_dexforce_w1_cfg(include_hand=False).to_dict()` and reading it back via `DexforceW1Cfg.from_dict()`, the left and right hands reappear; a config that builds only the left arm also gets the right arm back after the round-trip. I recommend modeling these options as first-class fields of `DexforceW1Cfg` and making `_build_defaults()` the single construction entry point; at minimum, sync `with_default_eef` and add a builder round-trip test.

embodichain/lab/sim/robots/dexforce_w1/params.py
```diff
@@ -210,17 +152,12 @@ def from_dict(cls, data: dict) -> "W1ArmKineParams":
             else data.get("arm_side")
         )
 
-        arm_kind = (
-            DexforceW1ArmKind[data["arm_kind"]]
-            if isinstance(data.get("arm_kind"), str)
-            else data.get("arm_kind")
-        )
         version = (
```
yuecideng: This parses by enum member name, so `"V025"` works, but `"v025"` — used by other config entry points and the docs — raises `KeyError`. I recommend providing a single `parse_w1_version()` (case policy, enum/value/name all handled in one place) shared by cfg, builder, spec, and kinematics, and covering `W1ArmKineParams.from_dict({"arm_side": "LEFT", "version": "v025"})`.

embodichain/lab/sim/robots/dexforce_w1/specs.py
```diff
@@ -0,0 +1,269 @@
...
+
+_W1_VERSION_SPECS = {
+    DexforceW1Version.V021: W1VersionSpec(
+        version=DexforceW1Version.V021,
+        component_urdfs=_V021_COMPONENT_URDFS,
+        full_robot_urdf_path="DexforceW1V021/DexforceW1_v02_1.urdf",
+        arm_d_list=_SHARED_D_LIST,
+        arm_base_z=0.1025,
+        default_eef_attach_xpos=_V021_DEFAULT_EEF_ATTACH_XPOS,
+        solver_tcp=_DEFAULT_TCP,
```
yuecideng, line 207: Although `W1VersionSpec` is `frozen=True`, the dicts passed in here are mutable, and V021/V022/V025 share the same `_DEFAULT_TCP` instance. If any caller mutates `spec.solver_tcp[...]`, it simultaneously pollutes the other versions and all subsequent configs, which contradicts the expectation of a "read-only single source of truth." I recommend using immutable mappings / fully tuple-ized data, or creating an independent defensive copy per spec and not exposing the internal containers directly.

embodichain/lab/sim/robots/dexforce_w1/utils.py
```diff
@@ -186,6 +166,14 @@ def __init__(self):
                 DexforceW1Version.V021,
             ): "DH_PGC_140_50_M/DH_PGC_140_50_M.urdf",
         }
+        for version in DexforceW1Version:
+            if version == DexforceW1Version.V021:
+                continue
+            for brand in DexforceW1HandBrand:
+                for side in DexforceW1ArmSide:
+                    self._urdf_rel_paths[(brand, side, version)] = self._urdf_rel_paths[
```
yuecideng: This loop silently maps every future W1 revision to the V021 hand asset: as soon as a new version is added to the enum, forgetting to confirm/register a hand asset will not fail-fast. At the same time, `DexforceW1Version` is used both as the whole-robot revision and as the external EEF version, coupling two separate dimensions. I recommend explicitly listing the supported aliases and raising on missing ones; longer term, `W1Revision` and `HandVersion/HandSpec` could be split apart.

embodichain/lab/sim/robots/dexforce_w1/cfg.py
```diff
@@ -66,129 +69,169 @@ def from_dict(
...
+            part_override = solver_cfg.get(part_name)
+            if not isinstance(part_override, dict) or "tcp" not in part_override:
+                continue
+            solver = self.solver_cfg.get(part_name)
+            if solver is None:
+                continue
+            spec = get_w1_version_spec(self._arm_version(arm_side))
+            solver.tcp = spec.compose_eef_attach_xpos(arm_side, solver.tcp)
+
+    def to_dict(self):
```
yuecideng, line 131: This introduces a W1-specific serialization protocol for derived state, but `agent_context/topics/robot-system/robot-system.md` — modified in this same PR — still states that a robot config's `from_dict` must keep the three-line template and `to_dict` must be inherited. The implementation and the maintenance rules now contradict each other, which makes incorrect refactoring easy down the line. I recommend at least syncing the agent context to spell out the W1 exception and the raw/final transform semantics; also add round-trip tests for V025 custom EEF/TCP, `component_versions`, and builder configs.

yuecideng (commented):
The overall direction looks good: centralizing version assets, kinematics, and EEF/TCP calibration into the spec layer makes the main path clearer. The line-by-line suggestions below focus on config round-trip, state consistency of the public builder, and fail-fast behavior when adding new versions/EEFs. The first two issues have already been reproduced on the current head; I recommend fixing them and adding regression tests before merging.

yuecideng (approved):

PR #457, opened 2026-08-05
feat(gen_sim): add Scene Engine and Gradio workspace

XuanchaoPENG:
# Description

This PR introduces an image-to-scene generation workflow for EmbodiChain, together with a Gradio-based workspace for running and previewing generative-simulation workflows.

Key changes:
- Add Scene Engine, which reconstructs tabletop scenes from an input image by performing scene understanding, segmentation, geometry generation, layout refinement, and scene export.
- Add embodichain scene-engine and embodichain preview-scene CLI commands.
- Add a Gradio workspace with managed pipeline execution and Viser scene previews.
- Add Scene Engine configuration through embodichain/gen_sim/.env and an example environment file.
- Add the scene-engine optional dependency group; extend the gensim extra with Scene Engine dependencies.

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have made corresponding changes to the documentation


embodichain/__main__.py
```diff
@@ -96,11 +96,6 @@ class Command:
         target="embodichain.workspace_cache_cli:main",
         help="Inspect and clean workspace analyzer caches.",
     ),
-    Command(
```
yuecideng, line 104: Recovery this command

embodichain/gen_sim/gradio_ui/gradio_app.py
```diff
@@ -0,0 +1,83 @@
...
+        raise FileNotFoundError(f"EmbodiChain root not found: {EMBODICHAIN_ROOT}")
+    demo = build_demo()
+    demo.queue(default_concurrency_limit=DEFAULT_CONCURRENCY_LIMIT)
+    _install_shutdown_handlers()
+    try:
+        demo.launch(
+            server_name=SERVER_NAME,
+            server_port=SERVER_PORT,
+            allowed_paths=[
+                str(EMBODICHAIN_ROOT),
```
yuecideng: Blocking security issue: this exposes the entire repository through Gradio while the server defaults to `0.0.0.0` and no authentication is configured. Gradio treats every file below an allowed directory as publicly servable, so this includes `embodichain/gen_sim/.env` and its API credentials. Please whitelist only the generated artifact/assets directories, explicitly block secret paths, and either bind to localhost by default or require authentication.

embodichain/gen_sim/gradio_ui/app_workflows.py
```diff
@@ -0,0 +1,3427 @@
...
+            return []
+        return remove_path(OUTPUTS_DIR)
+
+    errors: list[str] = []
+    for path in sorted(
+        OUTPUTS_DIR.rglob("*"),
+        key=lambda item: len(item.parts),
+        reverse=True,
+    ):
+        if path.is_file() and path.suffix.lower() not in VIDEO_SUFFIXES:
```
yuecideng: `OUTPUTS_DIR` is the repository-wide `outputs/` directory, so this loop deletes every non-video file below it whenever Reset or Auto cleanup runs. That includes unrelated RL checkpoints, debug reports, trajectories, and datasets. Please track and remove only artifacts created by the current Gradio run instead of recursively cleaning this shared directory.

embodichain/gen_sim/gradio_ui/app_media.py
```diff
@@ -0,0 +1,725 @@
...
+            continue
+        if mtime > latest_mtime:
+            latest_path = path
+            latest_mtime = mtime
+    return latest_path
+
+
+def configured_lerobot_roots() -> list[Path]:
+    roots: list[Path] = []
+    env_root = os.environ.get("EMBODICHAIN_DATASET_ROOT")
```
yuecideng: This function cannot run: `os` is not imported, and `CURRENT_PATHS` referenced below is not defined or imported by this module either. `monitor_simulation()` calls this path after DexSim exits but before clearing `runtime.sim_process`, so every completion raises `NameError` and leaves the UI/Auto loop stuck in a running state. Please import or pass these dependencies explicitly and make the runtime cleanup execute in a `finally` path.

yuecideng (commented):
Review focused on the new Gradio workspace. The inline comments cover security, data-loss, runtime, packaging, and CI blockers. Local validation also found that `pytest tests/test_main.py` fails because `analyze-workspace` is no longer registered, while `pytest tests/gen_sim` fails during collection because of the duplicate `test_config.py` module name. The existing inline thread on `embodichain/__main__.py` already calls out restoring that command, so I did not duplicate it.

embodichain/gen_sim/gradio_ui/app_articraft.py
```diff
@@ -0,0 +1,1138 @@
...
+        if listener_pids is None:
+            raise RuntimeError(
+                "Cannot identify the process using the Articraft Viser port."
+            )
+        if not listener_pids:
+            raise RuntimeError(
+                f"Port {self._port} is unavailable without a visible listener."
+            )
+
+        self._signal_listeners(listener_pids, signal.SIGTERM, "stop")
```
yuecideng: This treats every process listening on the configured port as a stale Articraft preview and sends it SIGTERM, later escalating to SIGKILL, without verifying ownership. A normal port collision can therefore terminate an unrelated user service. Please terminate only `self._process` (or another registered child owned by this app); otherwise report that the port is already in use.

embodichain/gen_sim/env.py
```diff
@@ -0,0 +1,105 @@
...
+        env: Environment mapping to populate. Defaults to :data:`os.environ`.
+
+    Returns:
+        The loaded path, or ``None`` when no local ``.env`` file exists.
+
+    Raises:
+        ValueError: If the file contains an invalid ``KEY=VALUE`` entry.
+    """
+    target_env = os.environ if env is None else env
+    env_path = find_gen_sim_env_file()
```
yuecideng, line 72: When neither `EMBODICHAIN_ENV_FILE` nor `embodichain/gen_sim/.env` exists, `find_gen_sim_env_file()` falls through and returns `None`, so this immediately raises `AttributeError: 'NoneType' object has no attribute 'is_file'`. A clean checkout only contains `.env.example`. Please return `Path | None` and guard `None` here, or always return the documented fallback path.

embodichain/gen_sim/gradio_ui/app_processes.py
```diff
@@ -0,0 +1,348 @@
...
+            start_new_session=True,
+            env=env,
+        )
+    )
+
+
+def build_pipeline_env() -> dict[str, str]:
+    env = os.environ.copy()
+    configure_direct_network_env(env)
+    configure_simready_llm_env(env)
```
yuecideng: `build_pipeline_env()` is used by Scene Engine, DexSim, Viser, and Articraft preview processes as well as SimReady, but this unconditionally maps `SIMREADY_OPENAI_*` over `OPENAI_*`. If separate endpoints are configured as supported by `.env.example`, Scene Engine will receive the SimReady model, URL, and key. Please apply this mapping only to the SimReady command.

embodichain/gen_sim/gradio_ui/app_workflows.py
```diff
@@ -0,0 +1,3427 @@
...
+            archive_run_log(
+                mode=RUN_LOG_MODE_AUTO,
+                task_description=task_text,
+                scene_description=scene_text,
+                outcome="stopped",
+            )
+            return "stopped"
+
+        with runtime_lock:
+            simulation_completed = (
```
yuecideng: `sim_returncode` is recorded by the monitor but never checked here. A non-zero DexSim exit still satisfies `sim_started && sim_finished && sim_process is None`, so Auto records the round as completed and continues to later phases. Please require `runtime.sim_returncode == 0` and also propagate a non-zero exit to the failed phase/`last_error`.

tests/gen_sim/scene_engine/test_config.py
```diff
@@ -0,0 +1,100 @@
...
+# ----------------------------------------------------------------------------
+
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import pytest
+
+from embodichain.gen_sim.scene_engine.cli import start
```
yuecideng, line 24: This new test module has the same basename as `tests/gen_sim/simready_pipeline/test_config.py`, and neither parent directory is a Python package. With the repository's default pytest import mode, `pytest tests/gen_sim` fails during collection with an import-file-mismatch error. Please rename this file (for example, `test_scene_engine_cli_config.py`) or make the test directories packages/use importlib mode.

embodichain/gen_sim/gradio_ui/app_asset_engine.py
```diff
@@ -0,0 +1,369 @@
...
+import shutil
+import subprocess
+import sys
+import threading
+import time
+import uuid
+from pathlib import Path
+from typing import Any, Iterable
+
+import gradio as gr
```
yuecideng, line 36: Gradio is now a direct runtime dependency of this feature, but neither the core dependency list nor the `gensim` extra declares it. A clean `pip install .[gensim]` therefore cannot reliably launch this UI. Please add a supported Gradio version to an appropriate optional extra and document that installation path.

.gitignore
```diff
@@ -208,3 +208,9 @@ scripts/benchmark/rl/reports/*
 .worktrees/
 # Local gym project workspace
 /gym_project/
+.debug_engine/
+
+# Local Gradio UI dependencies, generated Articraft records, and bytecode
```
yuecideng, line 213: Add Gradio dependancies

PR #468, opened 2026-08-07
update curobo sphere fit

matafela:
# Description

visualization: `motion_generator.planner.visualize_collision_models(control_part)` （only support curobo)

TODO:
- [x] curobo planner use dexsim sphere fit.
- [x] curobo world

## Type of change

## Checklist


embodichain/lab/sim/planners/curobo/curobo_planner.py
```diff
@@ -1624,6 +1599,39 @@ def _world_yaml_cache_key(self, world_cfg: CuroboWorldCfg) -> str:
             hasher.update(pose.detach().to("cpu").to(torch.float32).numpy().tobytes())
         return hasher.hexdigest()
 
+    def visualize_collision_models(
```
yuecideng: visualize_robot_collision_models would be better. Also this API should become base method of planner. For planner which does not support collision avoidance, we raise unsupported error.

embodichain/lab/sim/planners/curobo/curobo_planner.py
```diff
@@ -76,10 +75,16 @@
...
-# cached YAMLs from an older generator are regenerated instead of reused. v2:
-# exclude URDF mimic joints from cspace/lock_joints (cuRobo folds them into
-# their active joint and raises KeyError when locking one).
-_CUROBO_ROBOT_YAML_GENERATOR_VERSION = "v2"
+# cached YAMLs from an older generator are regenerated instead of reused. v2
+# excluded URDF mimic joints from cspace/lock_joints; v3 switched both robot
+# and obstacle fitting to DexSim MorphIt with fixed convex-hull limits; v4
+# removes self-collision metadata because the backend temporarily disables
+# cuRobo self-collision checking.
+_CUROBO_ROBOT_YAML_GENERATOR_VERSION = "v4-no-self-collision"
```
yuecideng: We can removed the version cache. Since it is not widely used now. Make it more maintainable in the future

embodichain/lab/sim/planners/curobo/curobo_planner.py
```diff
@@ -145,36 +150,26 @@ class CuroboWorldCfg:
...
+    voxel_size: float = 0.01
+    """ESDF voxel edge length in meters for every world collision object."""
 
-    ``"sphere"`` (default) fits spheres with cuRobo's
-    ``fit_spheres_to_mesh`` (fast collision queries, approximate, and requires
-    CUDA + cuRobo + trimesh). ``"cuboid"`` emits a local-frame AABB per object,
-    placed as an OBB via the object pose. ``"mesh"`` emits the full triangle
-    mesh (exact, no CUDA).
-    """
+    voxel_padding: float = 0.1
```
yuecideng: Is this param to large? How about curobo default?

yuecideng:
## Proposal: add an `auto` collision-representation policy based on DexSim physical collision shapes

The new single voxel-ESDF path is a useful simplification for a first implementation, but I do not think every scene object should necessarily be forced through voxelization. cuRobo V2 can keep cuboids, meshes, and voxel grids in the same `SceneData`, so the adapter can preserve a simple user API while choosing a more appropriate representation internally.

### Why the DexSim collision model should be the source of truth

The current world generator calls `RigidObject.get_vertices()` and `get_triangles()`. Those methods return the combined **visual mesh**, which can differ from the geometry used by physics. For example, EmbodiChain currently creates:

- `CubeCfg` as `RigidBodyShape.BOX`
- `SphereCfg` as `RigidBodyShape.SPHERE`
- regular mesh objects as `RigidBodyShape.CONVEX`
- ACD assets as multiple convex collision shapes
- meshes with `sdf_resolution > 0` as `RigidBodyShape.SDF`

DexSim already exposes the required runtime metadata on a rigid body:

- `get_shape_count()`
- `get_shape_type()`
- `get_shape_geometry(shape_idx)`
- primitive parameters such as box half-extents, sphere radius, and capsule radius/half-height
- convex/triangle-mesh vertices, triangles, and scale

`get_shape_type()` is an aggregate bitmask for compound bodies, so the adapter should not classify an object using a single equality check. It should enumerate `get_shape_count()` and inspect every shape geometry independently.

### Suggested `auto` mapping

| DexSim physical shape | Suggested cuRobo representation |
| --- | --- |
| `BOX` | `cuboid`; exact for the physics shape and normally the cheapest query |
| `PLANE` | a workspace-bounded thin cuboid |
| `SPHERE` / `CAPSULE` | preferably an analytic SDF checker; otherwise analytically generated voxel ESDF or a mesh fallback |
| `CONVEX` | the actual collision convex mesh |
| `MESH` | mesh for moderate geometry; voxel ESDF when mesh complexity and repeated-query cost justify it |
| `SDF` | reuse/convert to a cuRobo voxel grid when the grid data and conventions are compatible |
| compound / ACD | preserve the individual sub-shapes and build a mixed scene |
| unsupported/custom | explicit override or a documented conservative fallback |

Shape type should be the first decision key, not the only one. Mesh selection should also consider triangle count, static versus pose-dynamic behavior, expected query reuse, and an estimated voxel budget such as `prod(ceil(dims / voxel_size))`.

### Suggested public API

Keep the normal user path simple:

```python
CuroboWorldCfg(
    representation="auto",
    overrides={
        "room_scan": "voxel",
        "precision_fixture": "mesh",
    },
)
```

Internally, EmbodiChain could expose a planner-independent snapshot API such as:

```python
RigidObject.get_collision_shapes(env_id=0) -> list[CollisionShapeDesc]
```

This would avoid having the cuRobo adapter access private `obj._entities`, and it would ensure that planning uses the same collision geometry as DexSim physics. A compound object should map one object UID to multiple stable cuRobo obstacle names, with dynamic object pose updates fanned out through each shape local pose.

### DexSim gaps that should be addressed first

1. `ShapeGeometry.local_pose` is bound to Python, but the current `RigidBodyX::GetShapeGeometry()` implementation constructs geometry descriptors without copying `shape->getLocalPose()`. Compound and USD collision-shape offsets would therefore be lost.
2. `SDFGeometry` is bound, but the Python `get_shape_geometry()` dispatch does not include `SHAPE_SDF`, so an SDF can currently be detected through the type bitmask but its geometry cannot be retrieved through this API. Reusing an existing DexSim SDF would additionally require exposing its grid metadata/data or a canonical collision mesh.
3. The runtime geometry should be treated as already scaled; the adapter must avoid applying `RigidObjectCfg.body_scale` a second time.
4. Batched environments should validate that shape topology is identical across rows before cloning environment 0 geometry.

### Recommendation

I suggest keeping forced `voxel` as an explicit mode, but making `auto` the long-term default. In `auto`, simple physics primitives remain primitives, collision meshes remain meshes when practical, and ESDF is selected where it provides a real benefit. This preserves the simplified user experience without making voxel ESDF the universal scene representation, and it avoids planning against visual geometry that does not match the simulator collision model.

embodichain/lab/sim/planners/curobo/curobo_planner.py
matafela: It's a preprocess parameter for convex hull to voxel entry. I'll lower it to 0.005.

yuecideng (approved):

Issue #474, opened 2026-08-09
[Proposal] Simplify the atomic-action user API and unify runtime integration


yuecideng:
### Proposal

Review the atomic-action layer as a product-facing API, not only as a typed planning framework, and introduce a small semantic facade that makes common robot workflows usable without requiring callers to understand motion-planner selection, controllability filtering, grasp/EEF transforms, execution sessions, or effect-verifier plumbing.

This issue records the findings from the overall review of #450 and the atomic-action layer at commit `4ace46b64b9df476fe52e2c7578dc1e95fafacf8` on `feat/atomic-action-pr2b-dynamic-scene`.

The internal architecture is generally sound and should be preserved: typed goals/options, engine-owned planning resources, side-effect-free planning, immutable request snapshots, `ActionPlan`/`StateDelta`, explicit external-effect verification, and per-environment recovery. The main problem is that these internal concepts currently form the default user API.

#### Desired user experience

A normal pick-and-place workflow should be expressible at the semantic object level:

```python
skills = AtomicSkills.from_sim(sim, robot)
cube = skills.scene.object("cube")
tray = skills.scene.object("tray")

result = skills.run(
    Pick(cube),
    Place(cube, on=tray),
)
```

The default path should infer or configure once:

- arm/tool binding and open/grasp commands;
- grasp sampling, reachability, and controllability filtering;
- look-ahead constraints from later object goals;
- object-to-EEF pose conversion;
- planner/collision/recovery mode;
- planning context, observation loop, session, and runner;
- grasp/release/handover effect monitoring.

Only genuine ambiguity, such as choosing between two capable arms, should require an extra argument. Raw policies and planner protocols should remain available as advanced escape hatches.

#### Proposed API layers

1. **User layer**
   - `AtomicSkills` / `SkillRuntime`
   - semantic `Pick`, `Place`, and `HandOver` calls
   - `SceneObjectRef`
   - one synchronous/step-wise `SkillResult` contract

2. **Integration layer**
   - `RobotSkillProfile`
   - `SceneRegistry`
   - `GraspProvider`
   - `EffectMonitor`
   - named runtime/planning presets such as `safe`, `fast`, and `precise`

3. **Core/advanced layer**
   - the existing typed goals, invocations, bindings, plans, policies, contexts, sessions, runners, and provider protocols

The facade should compile to the existing core contracts. It should not introduce another independent execution framework.

#### P0 — cumulative sub-threshold scene movement is never published

[...]

Issue #471, opened 2026-08-08
[Proposal] Add declarative expert programs for atomic-action demonstrations


yuecideng:
### Proposal

Introduce a declarative **Expert Program** layer for generating expert demonstrations from the atomic-action system, and use it as the long-term replacement for Action Bank and its action-config format.

This layer should sit above the typed atomic-action contracts introduced by #448 and feed the segment-aware demonstration executor introduced by #460.

#### Proposed architecture

    ExpertProgramCfg / future MLLM SkillCallSpec
                      |
                      v
           InvocationCompiler / GoalGrounder
                      |
                      v
              ActionInvocation
                      |
              +-------+--------+
              |                |
              v                v
     engine.compile()   engine.start() + ExecutionSession.tick()
      static programs       observed / recoverable execution
              |                |
              +-------+--------+
                      |
                      v
               AtomicDemoBridge
          - PlanningContext provider
          - action effect verifier
          - command timing adapter
          - segment post-policies
                      |
                      v
              lazy DemoSegment(s)
                      |
                      v
           execute_demo_episode()

The configuration should compile to ActionInvocation objects instead of constructing action classes or calling the legacy engine.run() API. It should reuse the action registry and SkillDescriptor from #448 rather than introducing a second registry.

#### Configuration model

The initial schema should support:

- schema_version
- semantic robot-part bindings
- reusable skill presets containing:
  - skill_id
  - ActionBinding
  - ActionOptions
  - MotionPolicy
  - RecoveryPolicy
- registered object/semantic providers
- structured program nodes:
  - Sequence
  - Repeat
  - Segment
- registered goal grounders
- action-level effect verifiers
- segment post-policies, such as:
  - wait until a rigid body becomes stable
  - clear dynamics after a named phase
- task-level segment validators

Configuration must use strict, discriminated schemas and registered components. Arbitrary expressions such as eval("env...") should not be supported.

A later iteration can add Parallel, If, and Retry program nodes.

#### Runtime semantics

Use the two execution paths from #448 according to program semantics:

- engine.compile() for a static sequence whose later goals do not depend on observations produced by earlier actions.
- engine.start() and ExecutionSession.tick() when later goals require fresh observations, effect verification, or recovery.

The AtomicDemoBridge should adapt the atomic-action runtime to the environment generator used by demonstration collection:

1. capture the latest PlanningContext;
2. ground a semantic goal into an ActionInvocation;
3. run the execution session and emit environment commands;
[...]

PR #485, opened 2026-08-11
refactor(atomic-actions): establish snapshot identity bridge

yuecideng:
# Description

## Stack

- Layer: 1/28
- Base: `main`
- Depends on: none
- Followed by: https://github.com/DexForce/EmbodiChain/pull/486

Establish the Phase 0 snapshot/identity bridge used by every later declarative layer. Stable semantic entity IDs become authoritative for object grounding, action scene dependencies are explicit, object motion is derived from captured snapshots, and partial-batch state merges preserve verified identity deterministically.

The change deliberately fails closed for missing explicit IDs and keeps only the documented legacy no-ID fallback. It does not add a second planner, runtime, or scene source of truth.

Refs #471
Refs #474

## Type of change

- Breaking change (explicit semantic identities now fail closed instead of falling back to live state)

## Screenshots

Not applicable.

## Validation

- Focused coverage: `tests/sim/atomic_actions/test_actions.py`, `test_core.py`, and `test_engine_per_env.py`
- Final affected-suite regression on the stack tip: 1215 passed, 2 skipped, 8 deselected
- Changed Python files pass Black 26.3.1; the Sphinx build and rollout-report drift check pass at the stack tip

## Checklist

- [x] Changed Python files pass Black 26.3.1.
- [x] Corresponding public/design documentation is included in this stack.
- [x] Tests cover the affected behavior.
- [x] No dependency update is required.

### DIFF
--- a/agent_context/MAP.yaml
+++ b/agent_context/MAP.yaml
@@ -427,6 +427,8 @@ topics:
       - atomic actions
       - motion primitive
       - action primitive
+      - object semantics
+      - scene grounding
       - AtomicAction
       - ActionInvocation
       - AtomicActionEngine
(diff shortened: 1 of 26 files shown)


yuecideng:
Folded into #487 during stacked-PR consolidation. Its commits remain included in #487; the remote branch is retained for traceability.

PR #509, opened 2026-08-12
add articulation affordance

matafela:
# Description

- Add articulation affordance
- Add `twist` action: `python scripts/tutorials/atomic_action/twist.py (--rigid_object)`
- Add `press` action: `python scripts/tutorials/atomic_action/press.py (--rigid_object)`
- Add `slide` (for drawer, translation only): `python scripts/tutorials/atomic_action/slide.py`

## Type of change

## Checklist


yuecideng:
### Naming suggestion (non-blocking)

I suggest naming these primitives after the interaction they perform, while leaving target/category constraints to the affordance:

- `PressButton` → `Press` (skill ID: `press`), together with `PressGoal`, `PressOptions`, and `PressAffordance`. The implementation only relies on a press point, axis, and distance, and it already supports either an articulation link or a rigid object, so “button” is narrower than the actual behavior.
- `PullPushArticulatedPart` → `Slide` (skill ID: `slide`), together with `SlideGoal`, `SlideOptions`, and `SlideAffordance`. The invariant is grasping an articulation link and translating it along a configured axis. I would also prefer `direction: Literal["pull", "push"]` over `is_pull: bool`; if the articulation constraint must remain explicit in the class name, `SlideArticulatedPart` would be my second choice.
- `TurnKnob` → `Twist` (skill ID: `twist`), together with `TwistGoal`, `TwistOptions`, and `TwistAffordance`. This captures grasping and rotating about a specified axis, applies equally to knobs, caps, dials, and valves, and avoids implying arbitrary object reorientation as `Rotate` might.

For a threaded bottle cap, `Twist` accurately describes the current pure-rotation primitive. A future `Unscrew` action should remain separate if it explicitly plans pitch-driven axial translation or regrasping.

Overall, `Press` / `Slide` / `Twist` gives the public API a concise and consistent motion-centric vocabulary.

yuecideng (commented):
## Review findings

I reviewed the new affordance abstractions and the `PressButton`, `PullPushArticulatedPart`, and `TurnKnob` atomic actions against the atomic-action planning contract. I consider findings 1–4 blocking.

1. **[P1] The generated target frames are not guaranteed to be valid rotations.**

   `TurnAffordance.get_grasp_pose()` and `PressButtonAffordance.get_press_pose()` keep a fixed world-up Y axis and compute only `X = normalize(cross(Y, Z))`; Y is never recomputed. For an oblique axis such as `(1, 0, 1)`, the resulting matrix is not orthonormal (`det(R) ≈ 0.707`). The default `PressButtonAffordance.press_axis=(0, 0, 1)` with an identity target pose also fails because the cross product is zero. Please choose an adaptive reference axis, compute X, and then recompute `Y = cross(Z, X)`. Tests should assert orthonormality, determinant +1, and vertical-axis behavior.

2. **[P1] Press and pull/push motion is not constrained to the interaction axis.**

   `PressButton` plans only an approach endpoint and a pressed endpoint, without a contact waypoint or an axis-constrained Cartesian segment. `PullPushArticulatedPart` likewise supplies only a translated endpoint. With the default joint-space IK interpolation, the end effector may arc around the contact point or drawer axis. Please provide axis-aligned Cartesian keyframes/a constrained Cartesian segment and verify the FK path, not only the final pose.

3. **[P1] Replacing the stable `press` API is a breaking public API change.**

   This branch removes `Press`, `PressGoal`, `PressOptions`, the `press` module/tutorial/benchmark, and the stable `skill_id='press'`, replacing them with a non-equivalent `PressButton` API. Existing imports and plans will break, and the previous explicit-pose/`SceneEntityPose` targeting capability is lost. Please keep `Press` and add the new skill additively, or treat this as an explicit breaking release with deprecation and migration guidance. The atomic-action agent context also still lists `press` as a stable skill ID.

4. **[P1 / blocking] Affordances must not own live simulator entities.**

   Adding `Articulation`/`RigidObject` instances to affordances mixes a pure semantic/geometric capability description with mutable runtime binding. The actions then read live link poses directly instead of resolving state through `PlanningContext`/`SceneSnapshot`, which makes planning non-deterministic and bypasses dependency tracking. In particular, `collect_scene_dependencies(TurnKnobGoal(...))` is empty, so movement of the referenced object cannot trigger dynamic-goal recovery. Please keep affordances entity-free: store only local semantic/geometric data, put a stable entity/link reference in the goal or scene state, and resolve its pose through the planning context.

5. **[P2] The geometry contract does not describe the actual contact surface or rotation axis.**

   The default button contact point is the mean of all mesh vertices, which is commonly inside the object rather than on the pressable surface; the tutorial relies on this default. `TurnKnob` rotates around the link origin, but an axis direction alone does not define a 3D rotation axis, and the link origin need not lie on the physical joint axis. The affordances should explicitly model a surface contact point/region and an axis origin/pivot, ideally together with the associated joint/type/limits.

6. **[P2] The physical effects are neither declared nor verified.**

   All three new actions return an empty `StateDelta`. Consequently, trajectory completion can be reported as action success even if a drawer jams, a grasp slips, a knob does not turn, or a button is never actuated. Please add typed articulation/contact effects and terminal verification, or explicitly model and name these as open-loop motion primitives rather than completed semantic actions.

7. **[P2] Validation, export, and documentation regressions remain.**

   - `TurnKnobOptions(pre_grasp_distance=float('nan'))` is accepted; finite-value validation is missing.
   - `scripts/tutorials/atomic_action/tutorial_utils.py` exports the deleted `GRIPPER_TCP_Z`, so `from ...tutorial_utils import *` raises `AttributeError`.
   - The top-level atomic-action API autosummary omits `TurnKnob`, `TurnKnobOptions`, and `TurnAffordance`.

The existing atomic-action suite passes (`250 passed, 3 deselected`), but it currently does not cover the SO(3), Cartesian-path, scene-dependency, or physical-effect contracts above.

yuecideng (approved):

PR #548, opened 2026-08-24
Improve SRS IK arm-angle search and CPU/CUDA parity

chase6305:
# Description

This PR improves the correctness and performance of the 7-DoF SRS analytical IK solver.

The reference plane and seed redundancy are now calculated from the actual shoulder-elbow-wrist geometry. Seeded search starts from the seed’s geometric arm angle and expands outward, while full search covers the complete arm-
angle range.

CPU and CUDA implementations now use consistent arm-angle, periodic joint-distance, and joint-limit handling. The implementation also reduces repeated CPU geometry calculations, removes unnecessary CUDA combination buffers, reuses
Warp temporary arrays, and improves all-solution sorting.

Tests were added for geometric arm-angle sampling, periodic joint wrapping, runtime TCP/weight updates, and CPU/CUDA parity. A benchmark was also added for randomized, boundary, near-singular, and unreachable targets.

No new dependencies are required.

Fixes #N/A

## Type of change

- Bug fix
- Enhancement
- Documentation update

## Screenshots
```
python scripts/tutorials/sim/srs_solver.py --device cuda --num-steps 100
```
https://github.com/user-attachments/assets/b1aef048-027a-466b-9c1c-1643da069cd8

## Checklist

- [x] I have run the formatter on the changed files
- [x] Public API changes are reflected in the API docs, if applicable
- [x] I have added tests that prove my fix is effective
- [x] Dependencies have been updated, if applicable (no new dependencies)


embodichain/utils/warp/kinematics/srs_solver.py
```diff
@@ -339,6 +373,108 @@ def compute_fk_kernel(
...
+        return
+
+    elbow_model = dh_params[3 * 4 + 3] + qpos[tid * 7 + 3] * rotation_directions[3]
+    elbow_config = -1.0 if elbow_model < 0.0 else 1.0
+    shoulder_cosine = (
+        wp.pow(link_lengths[1], 2.0)
+        + wp.pow(distance, 2.0)
+        - wp.pow(link_lengths[2], 2.0)
+    ) / (2.0 * link_lengths[1] * distance)
+    q1_reference = wp.atan2(shoulder_to_wrist[1], shoulder_to_wrist[0])
```
yuecideng: **[P1] Use the same shoulder-azimuth degeneracy rule on CPU and CUDA**

This new CUDA seed-angle path always evaluates atan2(y, x), but `_calculate_arm_joint_angles` in both the CPU solver and the Warp reference-plane path sets q1 to zero whenever |z| <= 1e-6. A horizontal shoulder-to-wrist vector is not azimuth-degenerate; q1 is undefined only when hypot(x, y) is near zero. With seed [0, -pi/2, 0, -pi/2, 0, 0, 0], I get CPU/CUDA seed arm angles -2.18265 and -1.57080 (a 0.61186 rad gap), and neither path recovers the pose generated by its own FK. Please make all three paths use the same hypot(x, y) degeneracy check and add a horizontal-vector CPU/CUDA regression test.

embodichain/lab/sim/solvers/srs_solver.py
```diff
@@ -122,6 +132,93 @@ def _parse_params(self):
...
+    ) -> torch.Tensor:
+        """Remove periodic-equivalent solutions while preserving their order."""
+        if solutions.shape[0] < 2:
+            return solutions
+        pairwise_diff = solutions[:, None, :] - solutions[None, :, :]
+        wrapped_diff = (
+            torch.remainder(pairwise_diff + torch.pi, 2.0 * torch.pi) - torch.pi
+        )
+        equivalent = torch.amax(torch.abs(wrapped_diff), dim=-1) <= tolerance
+        has_equivalent_predecessor = torch.tril(equivalent, diagonal=-1).any(dim=1)
```
yuecideng: **[P1] Compare duplicates only with retained representatives**

`has_equivalent_predecessor` removes a row if it is close to any earlier input, even when that earlier row was itself removed. For [0, 0.9 * tolerance, 1.8 * tolerance] in one joint (all other joints equal), the function keeps only 0: the middle row is removed by the first, then the last is removed by the already-discarded middle row, despite being outside tolerance of the retained representative. Please use an order-preserving greedy/cluster selection against retained rows and add this chain-neighbor case. Also consider chunking or hashing, because `pairwise_diff` materializes N x N x 7 and makes `return_all_solutions` quadratic in GPU memory for large `num_samples`.

embodichain/utils/warp/kinematics/srs_solver.py
```diff
@@ -677,7 +833,8 @@ def sort_ik_kernel(
         dist = 0.0
         if valid:
             for j in range(7):
-                diff = qpos_out[idx * 7 + j] - qpos_seed[tid * 7 + j]
+                raw_diff = qpos_out[idx * 7 + j] - qpos_seed[tid * 7 + j]
+                diff = wp.atan2(wp.sin(raw_diff), wp.cos(raw_diff))
```
yuecideng: **[P3] Remove the obsolete Warp sorting kernel**

This PR no longer imports or launches `sort_ik_kernel`: all-solution ordering now uses `torch.argsort`, and a repository search finds only this definition. Keeping its O(N^2) selection-sort implementation (and updating its distance logic here) leaves a second, unreachable sorting path to maintain. Please remove the kernel.

yuecideng (commented):
Follow-up review with the three requested findings (items 1, 2, and 5).

yuecideng:
It would be better to add an example to demo this feature (may extend the existed SRS solver example)

chase6305:
> It would be better to add an example to demo this feature (may extend the existed SRS solver example)

yuecideng (approved):

PR #549, opened 2026-08-24
Enhance Pink IK Solver Robustness and Null-Space Posture Control

chase6305:
# Description

This PR improves the Pink IK solver and null-space posture task for more robust and consistent robot control.

Key changes include:

- Add batched IK with per-target seeds and consistent success reporting.
- Add adaptive damping, backtracking, stagnation detection, and strict convergence checks.
- Correct TCP and root-relative target transformations.
- Synchronize URDF, user-configured, and runtime robot joint limits.
- Fix null-space posture masking, tangent-space dimensions, joint ordering, and model caching.
- Include null-space posture objectives in the Pink QP and step acceptance flow.
- Improve handling of unconstrained task axes and invalid inputs.
- Add focused tests and update solver documentation.

No new runtime dependencies are introduced.

## Type of change

- Documentation update

## Screenshots

https://github.com/user-attachments/assets/61082743-b646-4f22-9336-25d9bb39ff11

## Checklist

- [x] I have run the black . command to format the code base.
- [x] Public API changes are reflected in the API docs.
- [x] I have added tests that prove my fix is effective.


embodichain/lab/sim/solvers/pink_solver.py
```diff
@@ -14,365 +14,625 @@
...
 
         Returns:
-            Target joint positions. (n_sample, 1, dof) of float.
+            Whether the limits were accepted.
         """
-        if qpos_seed is None:
-            qpos_seed = np.zeros(self.dof)
+        updated = super().set_qpos_limits(lower_qpos_limits, upper_qpos_limits)
+        if updated and hasattr(self, "_urdf_model_lower"):
+            self._sync_effective_limits()
```
yuecideng, line 364: **[P1] Preserve runtime robot limits across subsequent setter calls.**

`update_with_robot_limit()` folds the articulation bounds into `lower_qpos_limits` / `upper_qpos_limits`, but a later `set_qpos_limits()` call replaces those tensors before `_sync_effective_limits()` runs. This can therefore widen Pink beyond the robot's current hard limits. The new test currently demonstrates this by allowing the Pink-order lower bound for `joint1` to move from `-0.1` back to `-0.15`.

Please keep the URDF, user-configured, and runtime robot limits as separate sources and always synchronize their intersection. A regression test should cover both update orders and assert that no public setter can relax a runtime robot bound.

embodichain/lab/sim/solvers/pink_solver.py
```diff
@@ -14,365 +14,625 @@
...
+            lower_qpos_limits: Lower limit for every controlled joint.
+            upper_qpos_limits: Upper limit for every controlled joint.
 
         Returns:
-            Target joint positions. (n_sample, 1, dof) of float.
+            Whether the limits were accepted.
         """
-        if qpos_seed is None:
-            qpos_seed = np.zeros(self.dof)
+        updated = super().set_qpos_limits(lower_qpos_limits, upper_qpos_limits)
```
yuecideng: **[P2] Validate limit inputs transactionally before mutating solver state.**

`BaseSolver.set_qpos_limits()` does not validate the full vector shapes or finiteness, and it mutates the stored tensors before Pink synchronization. I reproduced a one-element input leaving `lower_qpos_limits` with shape `(1,)` before an `IndexError`, while NaN inputs return `True` and write NaNs into the Pinocchio model limits.

Please normalize both inputs first, require shape `(dof,)`, finite values, and element-wise `lower <= upper`, and only commit the new state after every check passes. The failure-path tests should also verify that the previous effective limits remain unchanged.

embodichain/lab/sim/solvers/pink_solver.py
```diff
@@ -14,365 +14,625 @@
...
+    def _task_metrics(self) -> tuple[float, float, float]:
+        """Return full task objective and frame-task convergence errors."""
+        objective = 0.0
+        position_error = 0.0
+        orientation_error = 0.0
+        for task in self.tasks:
+            error = np.asarray(task.compute_error(self.pink_cfg), dtype=float)
+            cost = 1.0 if task.cost is None else np.asarray(task.cost, dtype=float)
+            weighted = cost * float(task.gain) * error
+            objective += 0.5 * float(weighted @ weighted)
```
yuecideng: **[P1] Do not let an unprojected null-space error veto primary-task progress.**

This merit value includes the raw posture error `M e` for `NullSpacePostureTask`, while that task contributes the projected Jacobian `M N` to Pink's QP. When `N` is approximately zero, the posture task cannot influence the QP step, but moving toward the frame target can still increase its raw error and cause the later backtracking check to reject otherwise valid primary-task progress.

I reproduced this on SR5: the target succeeds without the posture task, but adding `NullSpacePostureTask(cost=10, controlled_frames=["ee_link"])` makes it fail even though the projector norm is about `1.5e-13`. The multiplication by `cost` itself is consistent with Pink; the issue is the mismatch between the raw-error merit and the projected task Jacobian.

Please use a primary-frame merit function (or lexicographic/hierarchical acceptance), and only use a controllable/projected posture term as a secondary criterion. A regression where the primary Jacobian is full rank and `N ~= 0` should show that adding this task does not change primary convergence.

embodichain/lab/sim/solvers/null_space_posture_task.py
```diff
@@ -147,19 +159,28 @@ def _build_joint_mapping(self, configuration: Configuration) -> None:
...
-            if joint_name in self.controlled_joints:
-                self._joint_mask[i] = 1.0
+        model = configuration.model
+        self._velocity_mask = np.zeros(model.nv)
+        available_joints = set(model.names.tolist()[1:])
+        unknown_joints = set(self.controlled_joints).difference(available_joints)
+        if unknown_joints:
+            raise ValueError(f"Unknown controlled joints: {sorted(unknown_joints)}")
+
+        selected_joints = set(self.controlled_joints) or available_joints
```
yuecideng, line 178: **[P2] Exclude floating-base coordinates from the default joint mask.**

For a FreeFlyer model, `available_joints` also contains the root joint. With an empty `controlled_joints` list, the default selection therefore marks all six root velocity coordinates as controlled, contrary to the docstring's statement that only actuated joints are affected. I reproduced a FreeFlyer-plus-one-revolute model producing a seven-element mask of all ones.

Please skip the root joint dimensions using Pink's root-joint dimension helper (or an equivalent Pinocchio-based calculation), while preserving explicit selection of actuated joints. A small floating-base regression test would make this public task's behavior unambiguous.

yuecideng (commented):
I left four inline comments covering the behavior and validation issues found during the review. The overall direction is good, but the two P1 items should be addressed before merging.

embodichain/lab/sim/solvers/pink_solver.py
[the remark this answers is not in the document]
chase6305: Thanks for raising this. Pink treats task.cost as a task-coordinate scaling factor rather than a direct quadratic coefficient. In Task.compute_qp_objective(), Pink constructs diag(cost), applies it to both the Jacobian and gain * error, and then forms the quadratic objective from the weighted quantities. Therefore, the corresponding residual objective is 0.5 * ||diag(cost) * gain * error||², which intentionally squares the configured costs. _task_metrics() matches that formulation, so using sqrt(cost) here would diverge from Pink's actual QP objective. No change is needed.

yuecideng (approved):

PR #554, opened 2026-08-26
add open door action

matafela:
# Description

- Add open door action.
- Add ur10-robotiq140 to tutorial: `python scripts/tutorials/atomic_action/coordinated_pickment.py --robot ur10 --auto_play`

open_door

## Type of change

## Checklist


embodichain/lab/sim/atomic_actions/primitives/open_door.py
```diff
@@ -0,0 +1,497 @@
...
+    door_waypoint_count: int = 20
+    """Number of Cartesian keyframes along the handle's circular arc."""
+
+    approach_distance: float = 0.1
+    """Pre-grasp distance opposite the automatically inferred approach axis."""
+
+    retract_distance: float = 0.1
+    """Post-release retreat distance opposite the rotated approach axis."""
+
+    open_angle: float = math.pi / 6
```
yuecideng: The requested hinge rotation is part of the semantic outcome, not merely a planner-tuning option. In its current signed form, this action can also close a door, while the planner does not resolve the current hinge position before applying the delta. Please move the desired opening state into `OpenDoorGoal` (for example, a target joint position/open fraction, or a positive opening angle plus an affordance-owned opening direction), evaluate it against the observed joint state and legal limits per environment, and fail rows whose target is invalid. If signed relative rotation is the intended contract, `RotateDoor` and `hinge_rotation_delta` would describe it more accurately than `OpenDoor` and `open_angle`.

embodichain/lab/sim/atomic_actions/affordance.py
```diff
@@ -364,6 +365,166 @@ def __post_init__(self) -> None:
...
+
+        current_link = link_name
+        visited: set[str] = set()
+        hinge = None
+        while current_link not in visited:
+            visited.add(current_link)
+            joint_info = joint_infos.get(current_link)
+            if joint_info is None:
+                break
+            joint_type = getattr(joint_info.joint_type, "name", joint_info.joint_type)
```
yuecideng: This traversal skips every non-revolute joint, although the public contract only says fixed intermediate joints are skipped. It can therefore cross a prismatic joint, or stop at a revolute handle/latch joint rather than the actual door hinge, producing an invalid opening arc. Please only auto-skip fixed joints and accept or require an explicit `hinge_joint_name` whenever the first active ancestor is ambiguous.

embodichain/lab/sim/atomic_actions/affordance.py
```diff
@@ -364,6 +365,166 @@ def __post_init__(self) -> None:
...
+            raise TypeError("link_name must be a string.")
+        if not link_name:
+            raise ValueError("link_name must be non-empty.")
+        link_names = tuple(articulation.link_names)
+        if link_name not in link_names:
+            raise ValueError(
+                f"Unknown articulation link {link_name!r}. Available links: "
+                f"{list(link_names)}."
+            )
+        entities = getattr(articulation, "_entities", None)
```
yuecideng: `from_articulation()` is a public factory, but it reaches into the private `_entities` collection and backend-specific joint-info objects. That couples this semantic affordance to articulation materialization and DexSim internals, so it cannot be used safely before binding. Please expose a narrow public articulation kinematic-chain query, or move this resolution into the simulation/scene-registry integration layer.

yuecideng (commented):
Focused review of the OpenDoor atomic-action contract and public API. I left inline comments on semantic naming, hinge resolution, backend coupling, the failing test configuration, and documentation/context coverage.

tests/sim/atomic_actions/test_actions.py
```diff
@@ -2404,6 +2424,235 @@ def position_fk(
...
+    context = _context()
+
+    plan = _plan_action(
+        action,
+        ActionInvocation(
+            skill_id="open_door",
+            goal=OpenDoorGoal(semantics, torch.eye(4)),
+            binding=_binding(action),
+            motion_policy=MotionPolicy(sample_count=18),
+            skill_options=OpenDoorOptions(hand_interp_steps=3),
```
yuecideng: This configuration fails CI before the test reaches the failed-row hold assertion: `sample_count=18` with `hand_interp_steps=3` and the default `door_waypoint_count=20` requires at least 33 samples. Since this test is about row-local grasp failure, set `door_waypoint_count=4` (minimum 17, preserving the expected trajectory length of 18), or raise `sample_count` and update the expectation. The normal defaults (`50/5/20`) are valid because `_motion_segment_lengths()` reserves 21 opening samples.

docs/source/overview/sim/atomic_actions/builtin_actions.md
```diff
@@ -673,6 +682,47 @@ axis-aligned Cartesian samples rather than sparse joint-space endpoints.
...
+hinge-to-handle radial vector, then samples a handle grasp as `Slide` does. For
+the opening segment it interpolates handle-link poses around the resolved
+hinge axis and applies the initial rigid `link -> EEF` transform to recover the
+corresponding EEF poses. After release, retract follows the approach direction
+after it has rotated with the open door.
+
+| Contract | Value |
+|---|---|
+| Skill ID | `open_door` |
+| Goal | `OpenDoorGoal(semantics=..., target_pose=...)` |
```
yuecideng: `OpenDoorGoal.target_pose` accepts `SceneEntityPose`, and the plan monitors that dependency through the `reach` segment, but the shared dynamic-pose table above does not include this action. Please add the corresponding row and update `agent_context/topics/atomic-actions/atomic-actions.md` (and `agent_context/MAP.yaml` keywords if needed), since this PR changes the registered atomic-actions source of truth.

embodichain/lab/sim/atomic_actions/affordance.py
matafela: Add `get_parent_joint_chain` to `Articulation`

yuecideng (approved):

PR #566, opened 2026-08-30
Add batched trapezoidal and Double-S trajectory planning

chase6305:
# Description

  This PR adds a batched trajectory planner supporting trapezoidal velocity and jerk-limited seven-phase Double-S motion profiles.

  The planner supports both Torch and NVIDIA Warp backends and integrates with the existing MotionGenerator interface. The Warp backend parallelizes scalar profile construction, phase evaluation, and batched joint trajectory
  composition.

  It also adds Cartesian straight-line trajectory planning for the simulation tutorial. Cartesian distance is time-parameterized before IK, ensuring the requested end-effector path follows the configured velocity, acceleration, and
  jerk constraints.

  For OPW-based robots, the complete Cartesian pose path is now solved through a batched Warp path-IK interface:

  1. One Warp launch computes all analytical OPW candidates for every Cartesian sample.
  2. A second Warp launch selects a temporally continuous IK branch for each environment.
  3. Each sample uses the previously selected joint state as its seed.
  4. Joint-limit-valid 2π equivalent configurations are handled during branch selection.

  This avoids launching OPW IK separately for every trajectory sample while preserving joint-path continuity.

  Additional changes include:

  - Add TrapezoidalPlanner, TrapezoidalPlannerCfg, and TrapezoidalPlanOptions.
  - Support scalar and per-joint velocity, acceleration, and jerk limits.
  - Support synchronized multi-joint motion.
  - Support triangular fallback for short trapezoidal moves.
  - Support seven-phase Double-S profiles without display filtering.
  - Support fixed-count and fixed-time trajectory sampling.
  - Support optional minimum-duration scaling.
  - Support redundant collinear waypoint compression.
  - Add explicit torch, warp, and auto backend selection.
  - Add continuous Cartesian path IK through Robot.compute_ik_path().
  - Add batched OPW path solving through OPWSolver.get_ik_path().
  - Compute Cartesian joint velocity and acceleration using the path time law and differential kinematics rather than numerical time differentiation.
  - Add simulation replay using the planned trajectory timing.
  - Add diagnostic plots for:
      - End-effector XYZ path
      - End-effector orientation
      - Joint position
      - Joint velocity
      - Joint acceleration

  - Add standalone profile and simulation tutorials.
  - Add a reproducible Torch/Warp trajectory-planner benchmark.
  - Update motion-planning, solver, and API documentation.

  Motivation and context:

  The existing trajectory-planning examples primarily relied on TOPPRA and did not provide a lightweight, natively batched implementation for trapezoidal or jerk-limited motion profiles.

  Cartesian trajectories also require the time law to be applied in Cartesian space before IK. Applying joint-space timing after independently sampled IK can distort the requested end-effector path and produce inconsistent
  derivatives.

  This change provides a maintainable trajectory-planning implementation suitable for both CPU execution and large batched CUDA workloads, while keeping Torch as the reference and fallback backend.

  Dependencies:

  - No new runtime dependencies.
  - NVIDIA Warp is already used by EmbodiChain and is required when explicitly selecting backend="warp".
  - Matplotlib is used by the existing tutorial visualization environment.

  Fixes #<issue-number>

  ## Type of change

  - [x] Documentation update

  ## Screenshots

• ## Examples

  The examples below cover single-trajectory planning, batched planning, full CobotMagic integration, numerical validation, and plot generation.

  ### Single Cartesian line trajectory

  Run a jerk-limited Double-S Cartesian line trajectory for one CobotMagic environment:
```bash
  python scripts/tutorials/sim/planner/trapezoidal_planner.py \
    --num_envs 1 \
    --path cartesian \
    --cartesian-path line \
    --profile acceleration_trapezoidal \
    --samples 300 \
    --plot-output outputs/cobotmagic_single_line.png
```

https://github.com/user-attachments/assets/80cf5a58-d1ad-40be-aea8-2a9cd2efaa54

  ### Single Cartesian Bézier trajectory
```bash
  python scripts/tutorials/sim/planner/trapezoidal_planner.py \
    --num_envs 1 \
    --path cartesian \
    --cartesian-path bezier \
    --profile acceleration_trapezoidal \
    --samples 300 \
    --plot-output outputs/cobotmagic_single_bezier.png
```

https://github.com/user-attachments/assets/0f455c48-faf4-43a8-a579-090ec6c3e954

  ### Single multi-waypoint joint trajectory

  Run a joint-space trajectory with jerk-limited timing and quintic corner blending:
```bash
  python scripts/tutorials/sim/planner/trapezoidal_planner.py \
    --num_envs 1 \
    --path joint \
    --profile acceleration_trapezoidal \
    --blend-tolerance 0.05 \
    --samples 300 \
    --plot-output outputs/cobotmagic_single_joint_blend.png
```

https://github.com/user-attachments/assets/e1ec05ea-fb77-44e3-8025-03d9ebf44bd0

  ### Batched Cartesian trajectories

  Plan Cartesian line trajectories for four environments:
```bash
  python scripts/tutorials/sim/planner/trapezoidal_planner.py \
    --num_envs 4 \
    --path cartesian \
    --cartesian-path line \
    --profile acceleration_trapezoidal \
    --samples 300 \
    --plot-env 0 \
    --plot-output outputs/cobotmagic_batch_line.png
```
https://github.com/user-attachments/assets/7bc1db88-7034-4e29-97f1-2036fab6e54a

  ### Batched multi-waypoint joint trajectories
```bash
  python scripts/tutorials/sim/planner/trapezoidal_planner.py \
    --num_envs 4 \
    --path joint \
    --profile acceleration_trapezoidal \
    --blend-tolerance 0.05 \
    --samples 300 \
    --plot-env 0 \
    --plot-output outputs/cobotmagic_batch_joint_blend.png
```
  ### Full CobotMagic integration

  Run both joint-space and Cartesian trajectories with both supported time profiles:
```bash
  python scripts/tutorials/sim/planner/trapezoidal_planner.py \
    --num_envs 4 \
    --path both \
    --profile both \
    --cartesian-path line \
    --blend-tolerance 0.05 \
    --minimum-duration 3.0 \
    --samples 300 \
    --plot-env 0 \
    --plot-output outputs/cobotmagic_full_line.png
```
  This exercises:

  - Joint-space trapezoidal planning
  - Joint-space jerk-limited Double-S planning
  - Cartesian line planning with continuous IK
  - Batched trajectory generation
  - Quintic joint-waypoint blending
  - Minimum-duration scaling
  - Diagnostic plotting
  - Simulator trajectory replay

  To exercise the Cartesian Bézier path instead, replace:

  --cartesian-path line

  with:

  --cartesian-path bezier

  For headless execution, add:

  --headless --no-show-plot

  For example:
```bash
  python scripts/tutorials/sim/planner/trapezoidal_planner.py \
    --headless \
    --no-show-plot \
    --num_envs 8 \
    --path both \
    --profile both \
    --cartesian-path line \
    --blend-tolerance 0.05 \
    --minimum-duration 3.0 \
    --samples 300 \
    --plot-output outputs/cobotmagic_headless.png
```
  ## Simulation-free numerical validation

  Run all numerical scenarios without starting the simulator:
```
  for scenario in \
    bezier \
    trapezoidal \
    double-s \
    blend \
    minimum-duration \
    batch \
    se3 \
    backend
  do
    python scripts/tutorials/sim/planner/trajectory_pr_checks.py "$scenario" || exit 1
  done
```
  The scenarios cover:

  - Quadratic and quintic Bézier evaluation
  - Exact Bézier derivatives and arc length
  - Trapezoidal timing
  - Seven-phase Double-S timing
  - Multi-waypoint quintic blending
  - Minimum-duration scaling
  - Mixed moving/stationary batches
  - Cartesian SE(3) screw interpolation
  - Torch/Warp backend consistency
  - Realized velocity, acceleration, and jerk reports

  Each scenario prints [PASS] when all invariants hold and exits with a non-zero status on failure.

  ## Plot generation

  Generate all simulation-free diagnostic plots:
```
  for scenario in \
    bezier \
    trapezoidal \
    double-s \
    blend \
    minimum-duration \
    se3
  do
    python scripts/tutorials/sim/planner/trajectory_pr_plots.py "$scenario" || exit 1
  done
```
  Plots are written to:
```
  outputs/trajectory_plots/
```
  Generate and display one plot interactively:
```bash
  python scripts/tutorials/sim/planner/trajectory_pr_plots.py blend \
    --output outputs/trajectory_plots/blend.png \
    --show
```
  ## Focused tests
```bash
  pytest -q \
    tests/sim/planners/test_bezier.py \
    tests/sim/planners/test_trapezoidal_planner.py \
    tests/sim/planners/test_se3.py \
    tests/sim/planners/test_motion_generator.py \
    tests/sim/planners/test_motion_generator_batched.py
```
  ## Default-planner direction

  The reusable implementation is integrated through TrapezoidalPlanner and TrapezoidalPlannerCfg; the tutorial scripts are not production planner entry points.

  For training and demonstration collection, the intended production defaults are:
```
  TrapezoidalPlanOptions(
      profile="double_s",
      stop_at_waypoints=False,
      blend_tolerance=0.0,
      backend="auto",
  )
```
  This provides bounded jerk and deterministic batching while preserving exact waypoints. Quintic geometric blending remains opt-in because a non-zero blend tolerance intentionally permits waypoint deviation.

  
  ## Validation

  The following focused validation was completed:

  ```text
  57 passed

  Covered behavior includes:

  - Torch and Warp scalar-profile construction consistency.
  - Torch and Warp trajectory sampling consistency.
  - Trapezoidal and Double-S boundary conditions.
  - Stationary and short-distance segments.
  - Velocity, acceleration, and jerk constraint enforcement.
  - Cartesian straight-line geometry.
  - Cartesian time-law propagation.
  - Continuous OPW Warp branch selection.
  - Batched tutorial planning and plotting behavior.
  - Python compilation and formatting of affected files.
  - git diff --check.

  The OPW candidate and continuous-path selection kernels were compiled and tested with the CPU Warp backend. CUDA uses the same Warp kernels, but full GPU simulation validation should also run in CI or on a machine with an
  available CUDA driver.

  ## Checklist

  - [x] I have run the black . command to format the code base.
  - [x] I have formatted all affected Python files with Black.
  - [ ] Public API changes are reflected in the API docs (python docs/scripts/check_api_docs.py), if applicable.
  - [x] Dependencies have been reviewed; no dependency updates are required.


scripts/tutorials/sim/planner/trapezoidal_planner.py
```diff
@@ -0,0 +1,854 @@
+# ----------------------------------------------------------------------------
```
yuecideng, line 1: Recommended to add a new folder to put all planner tutorial together or move these two files to examples/sim/planner

yuecideng:
## Review against the current `origin/main`

Thanks for the substantial work in this PR. The new batched joint-space time-parameterization backend is useful, and the `BasePlanner`/`MotionGenerator` registration, explicit `PlanResult.dt`, and Torch/Warp split are directionally consistent with the existing motion-planning stack.

I reviewed PR head `0a2752bd` against the current `origin/main` (`aeca62f`). The synthetic merge is clean, but I recommend **requesting changes before merge** because the current implementation has one P0 architecture blocker and three reproducible P1 trajectory-correctness issues.

### What this PR adds

- `TrapezoidalPlanner`, `TrapezoidalPlannerCfg`, and `TrapezoidalPlanOptions`.
- Batched triangular/trapezoidal and seven-phase Double-S profiles.
- Scalar or per-joint velocity, acceleration, and jerk constraints.
- Torch reference implementation plus a float32 Warp backend for profile construction and sampling.
- Fixed-count/fixed-time sampling, minimum-duration scaling, and optional waypoint compression.
- A Cartesian straight-line tutorial using a Cartesian scalar time law, batched IK, differential kinematics, plotting, and replay.
- Tutorials, a benchmark, tests, API docs, and agent context updates.

## Blocking findings

### [P0] Reuse `compute_batch_ik`; do not add a second Path IK stack

The existing [`Robot.compute_batch_ik`](https://github.com/DexForce/EmbodiChain/blob/aeca62f98feb10543e44f9c688cf7d7b83ee1f4e/embodichain/lab/sim/objects/robot.py#L1014-L1112) already owns the batch-IK boundary:

- accepts `(B, N, 4, 4)` or `(B, N, 7)` pose batches;
- converts arena-frame targets into the solver root frame;
- flattens `(B, N)` into one solver batch;
- calls `solver.get_ik()` once; and
- restores `(B, N, DOF)` results.

The existing [`OPWSolver.get_ik_warp`](https://github.com/DexForce/EmbodiChain/blob/aeca62f98feb10543e44f9c688cf7d7b83ee1f4e/embodichain/lab/sim/solvers/opw_solver.py#L229-L368) also already computes all eight OPW candidates for an arbitrary target batch in one Warp launch and supports `return_all_solutions=True`.

The new [`Robot.compute_ik_path`](https://github.com/DexForce/EmbodiChain/blob/0a2752bd0fd214ac6dec4f0b5f17b4622b6ed711/embodichain/lab/sim/objects/robot.py#L1083-L1127) and [`OPWSolver.get_ik_path`](https://github.com/DexForce/EmbodiChain/blob/0a2752bd0fd214ac6dec4f0b5f17b4622b6ed711/embodichain/lab/sim/solvers/opw_solver.py#L375-L461) duplicate coordinate conversion, validation, flattening/reshaping, OPW candidate allocation, limit/TCP preparation, and the candidate kernel launch. The only genuinely new behavior is sequential candidate selection using the previous sample as the next seed.

Requested change:

- Remove the new public `Robot.compute_ik_path()` and `OPWSolver.get_ik_path()` APIs.
- Keep `Robot.compute_batch_ik()` as the single batch-IK entry point.
- Extend that existing path (or its existing solver options) with an optional continuous-selection mode.
- Reuse `solver.get_ik(..., return_all_solutions=True)` for OPW candidate generation.
- Keep the sequential candidate selector as an internal implementation detail; it must not launch a second, duplicated OPW candidate solver.
- Update the tutorial, tests, API docs, and agent context to use the existing batch-IK entry point.

Calling the current `compute_batch_ik()` unchanged is not quite equivalent because it hardcodes `return_all_solutions=False` and independently selects each target. A small extension is therefore needed to preserve the PR's claimed branch-continuity behavior, but a parallel public Path IK API is not needed.

### [P1] Multi-segment sampling drops required internal waypoints

[`_make_sample_times`](https://github.com/DexForce/EmbodiChain/blob/0a2752bd0fd214ac6dec4f0b5f17b4622b6ed711/embodichain/lab/sim/planners/trapezoidal_planner.py#L590-L610) samples only a global grid over total duration. Segment boundary times are not inserted before composing the output trajectory.

For `A=(0,0) -> B=(1,0) -> C=(1,1)`:

- `sample_count=2` returns only `[A, C]`;
- `sample_count=4` returns approximately `A, (0.68,0), (1,0.32), C`; and
- even `sample_count=10` never contains B exactly.

The continuous internal profile is not part of `PlanResult`; consumers only receive the sampled positions. Consequently, `stop_at_waypoints=True` does not ensure that execution reaches or stops at the waypoint, and a controller can cut across a required corner.

Please build sample times per segment so every retained cumulative segment boundary appears exactly in the output. In quantity mode, reject requests with fewer samples than retained waypoints, then allocate remaining samples across segments. Add a regression asserting exact inclusion of B and zero velocity at B.

### [P1] Duplicate waypoints can delete a genuine corner

The current compression rule removes any interior point adjacent to a zero-length edge: [`duplicate_neighbor = ...`](https://github.com/DexForce/EmbodiChain/blob/0a2752bd0fd214ac6dec4f0b5f17b4622b6ed711/embodichain/lab/sim/planners/trapezoidal_planner.py#L148-L181).

Reproduction:

```text
[A=(0,0), B=(1,0), B=(1,0), C=(1,1)] -> [A, C]
```

This changes an L-shaped path into a diagonal whenever `stop_at_waypoints=False`.

Please collapse each consecutive duplicate run first while preserving one representative of the corner, and then test collinearity against the nearest non-duplicate neighbors. Add the exact `A, B, B, C` regression case.

### [P1] Explicit motion limits are silently ignored when `sample_count=None`

The trapezoidal option construction is incorrectly guarded by `sample_count is not None` in [`MotionGenerator.resolve_plan_options`](https://github.com/DexForce/EmbodiChain/blob/0a2752bd0fd214ac6dec4f0b5f17b4622b6ed711/embodichain/lab/sim/planners/motion_generator.py#L391-L441).

This call:

```text
sample_count=None, velocity_limit=0.01, acceleration_limit=0.02
```

returns the backend defaults `velocity=0.2` and `acceleration=0.5`. A caller's explicit limits can therefore be silently relaxed.

Please resolve backend options independently of sampling: always apply non-`None` generic limits, and let `sample_count` control only the sampling method/count. Add a `MotionGenerator`-level regression test rather than testing only `TrapezoidalPlanner.__new__()`.

## Additional architecture and implementation improvements

### [P2] Cartesian planning is implemented as a tutorial-private pipeline

The core `TrapezoidalPlanner` supports only `JOINT_MOVE`, while the tutorial imports the private [`_plan_linear_profiles`](https://github.com/DexForce/EmbodiChain/blob/0a2752bd0fd214ac6dec4f0b5f17b4622b6ed711/scripts/tutorials/sim/trapezoidal_planner.py#L37-L47) helper and independently assembles the Cartesian path, IK, derivatives, and replay.

If Cartesian-first timing is intended to be a reusable product capability, expose it through a typed public planning contract composed by `MotionGenerator`. If it is intentionally tutorial-only, narrow the PR/docs claims accordingly and avoid describing it as a generally available planner feature.

### [P2] Batched replay does not preserve per-environment timing

[`replay_plan`](https://github.com/DexForce/EmbodiChain/blob/0a2752bd0fd214ac6dec4f0b5f17b4622b6ed711/scripts/tutorials/sim/trapezoidal_planner.py#L418-L445) advances every environment using `dt[:, sample_index].max()`.

When batch rows have different durations, only the longest row follows its planned timing. Shorter rows are slowed down, so their executed velocity/acceleration no longer match `PlanResult`.

Please resample all rows onto one shared simulation-time grid before replay, or explicitly restrict replay to `B=1`/identical `dt` rows and validate that precondition.

### [P2] Public integration coverage is incomplete

The current tests cover the scalar planner, tutorial helpers/fakes, and the OPW selector kernel, but they do not exercise the real public integration path end to end.

Please add focused coverage for:

- `MotionGenerator` construction and generation with `TrapezoidalPlannerCfg`;
- `Robot.compute_batch_ik()` with a `(B,N)` OPW pose path and continuous selection;
- OPW FK -> batch IK reconstruction with TCP/base transforms;
- configured `ik_nearest_weight` during continuous selection;
- the three regressions above; and
- Torch/CUDA-Warp parity on an available GPU CI worker.

The motion-planning agent context should also stay focused on ownership, entry points, contracts, and validation surfaces. Tutorial font, dashboard layout, and plotting details belong in tutorial/Sphinx documentation rather than agent architecture context.

## Validation performed

I merged the PR into the current `origin/main` in a temporary worktree and ran the new planner/tutorial/selector tests together with existing MotionGenerator, OPW solver, and agent-context regressions:

```text
96 passed, 2 skipped
```

Additional checks passed:

- API docs checker: `1679/1679` public exports documented.
- Black check on the changed Python files.
- `git diff --check`.

The GitHub lint/build/test checks are green on the PR head, but they predate the latest `origin/main` changes. Full CUDA Warp and real simulation replay validation are still outstanding.

## Recommendation

Please address the P0 reuse issue and all three P1 correctness issues before merge. The P2 items should either be resolved in this PR or explicitly scoped/documented so the public API does not promise behavior that currently exists only in the tutorial.

embodichain/utils/warp/kinematics/trapezoidal_warp.py
```diff
@@ -0,0 +1,399 @@
+# ----------------------------------------------------------------------------
```
yuecideng, line 1: For kernel-only module, we should put them into `embodichain/utils/warp/kinematics`

yuecideng (changes requested):

scripts/tutorials/sim/planner/trapezoidal_profile.py
```diff
@@ -0,0 +1,134 @@
+# ----------------------------------------------------------------------------
```
yuecideng, line 1: Should we standardlize this script as a reusable tool for motion generator @chase6305 ?

yuecideng (approved):

scripts/tutorials/sim/planner/trapezoidal_profile.py
chase6305: @yuecideng 
This is already supported through the reusable TrapezoidalPlanner backend registered with MotionGenerator.
```python
motion_generator = MotionGenerator(
     MotionGenCfg(
         planner_cfg=TrapezoidalPlannerCfg(robot_uid=robot.uid),
     )
)
````
Callers can select either the trapezoidal or jerk-limited Double-S profile through TrapezoidalPlanOptions. The script is intentionally kept as a lightweight, simulation-free visualization example, so I would not promote the script itself into a production utility.
We may follow up by making the Double-S backend the production fallback for training and demonstration collection, while keeping TOPPRA explicitly selectable for time-optimal use cases.

File: embodichain/lab/sim/planners/trapezoidal_planner.py [hunk 4]
```diff
@@ -0,0 +1,1125 @@
...
+    velocities = velocities / derivative_scale[:, None, None]
+    accelerations = accelerations / derivative_scale[:, None, None].square()
+    joint_jerks = joint_jerks / derivative_scale[:, None, None].pow(3)
+    if options.profile == "double_s":
+        peak_jerk_joint = joint_jerks.abs().amax(dim=1)
+        peak_jerk = peak_jerk_joint.amax(dim=1)
+        # A uniform time stretch scales velocity, acceleration, and jerk by
+        # 1/k, 1/k², and 1/k³. Correct the curvature-variation contribution
+        # that is not represented by the tangential jerk projection.
+        jerk_scale = (
```
yuecideng, line 849: **[P1] Make blended constraint enforcement independent of output sampling**

The correction here is computed only from the returned `velocities`, `accelerations`, and `joint_jerks`, whose density is controlled by `sample_interval`. For `waypoints=[[[0, 0], [1, 0], [1, 1]]]`, velocity limits `[0.8, 0.6]`, acceleration limits `[1.0, 0.7]`, jerk limits `[2.0, 1.5]`, `profile="double_s"`, and `blend_tolerance=0.1`, using `sample_interval=2` returns duration `5.626`, reports peak jerk `1.456`, and sets `within_limits=True`; a 10,001-point evaluation of that same continuous profile reaches approximately `4.272` on each joint. Raising only the output resolution to 4,001 samples stretches the duration to `7.970`. Thus the same path and limits receive a different safety result solely because of output resolution. Please compute the correction on a resolution-independent internal grid or from an analytic derivative bound before producing caller-requested samples, and add a low-vs-high sampling regression test.

embodichain/lab/sim/planners/utils.py
```diff
@@ -209,6 +209,9 @@ class PlanResult:
     dt: torch.Tensor | None = None
     """Per-env time deltas, shape ``(B, N)``."""
 
+    constraint_report: dict[str, torch.Tensor] | None = None
```
yuecideng, line 212: **[P2] Preserve `constraint_report` through `MotionGenerator`**

Adding the field here does not make it observable through the project planning facade: `MotionGenerator._normalize_plan_result()` reconstructs `PlanResult` at `motion_generator.py:905-912` without copying `constraint_report`, so `MotionGenerator.generate()` always drops the planner report even when no resampling occurred. Please propagate `result.constraint_report` when the trajectory is unchanged; if resampling occurs, either recompute it or explicitly invalidate it. A focused test should exercise this through `MotionGenerator.generate()`, rather than only calling `TrapezoidalPlanner` directly.

yuecideng (commented):
Focused follow-up on the latest head. The three actionable findings and one architecture suggestion are attached inline.

embodichain/lab/sim/planners/se3.py
```diff
@@ -0,0 +1,301 @@
...
+
+    rho, rotation = _se3_log(torch.linalg.inv(start) @ end)
+    tangent = torch.cat((rho, rotation))
+    active = tangent.abs() > torch.finfo(start.dtype).eps
+    if not bool(active.any().item()):
+        raise ValueError("start and end must describe distinct poses.")
+    scalar_limits = torch.stack(
+        [(limit[active] / tangent[active].abs()).amin() for limit in limits]
+    )
+    scalar_profile = _build_scalar_profile(
```
yuecideng, line 248: **[P2] Validate SE(3) timing options at the public boundary**

`Literal` does not enforce runtime inputs here. `_build_scalar_profile()` treats every name other than `"double_s"` as trapezoidal, while its 1% safety margin is applied only to recognized names; `_apply_minimum_duration()` also silently ignores a negative duration. Consequently, `plan_se3_line(profile="invalid")` returns a plausible trajectory with different semantics instead of rejecting the input. Please validate `profile` and require a finite, non-negative `minimum_duration` at this public boundary. Since this module also imports four private helpers from `trapezoidal_planner`, please extract a shared typed scalar-time-law abstraction rather than coupling one public planner utility to another module's private implementation.

embodichain/lab/sim/objects/robot.py
```diff
@@ -1080,14 +1107,31 @@ def compute_batch_ik(
         )
         pose_batch = torch.bmm(base_inv_xpos_batch, pose_batch)
 
+        if continuous:
+            candidate_valid, candidate_qpos = solver.get_ik(
+                target_xpos=pose_batch,
+                qpos_seed=None,
+                return_all_solutions=True,
+            )
+            select_path = getattr(solver, "_select_continuous_ik_path", None)
```
yuecideng, line 1116: **Suggestion for this PR: declare continuous batch IK as a typed `BaseSolver` capability**

`Robot` currently discovers solver behavior with `getattr()` on a private method, and it does so only after generating all IK candidates. Please move this optional capability contract into `BaseSolver` in this PR—for example, a typed support property plus a protected selection method with a default unsupported implementation—then override it in `OPWSolver` and check support before calling `get_ik(return_all_solutions=True)`. This keeps the robot layer solver-agnostic, provides static type coverage, and avoids wasted candidate generation or unsupported-call failures for other solvers.

embodichain/utils/warp/kinematics/opw_solver.py
```diff
@@ -14,8 +14,6 @@
 # limitations under the License.
 # ----------------------------------------------------------------------------
```
yuecideng, line 16: Return

embodichain/utils/warp/kinematics/trapezoidal_warp.py
yuecideng, line 1: Move to conpute/kinematics module

yuecideng (changes requested):

yuecideng (approved):

PR #573, opened 2026-09-01
Articulation atom action axis auto generation.

matafela:
# Description

- `translation_axis` of `Slide` is now auto generated.
- `twist_axis`, `axis_origin` of `Twist` is now auto generated.
- `press_axis` of `Press` is now auto generated.
- Fix tutorial `scripts/tutorials/atomic_action/pour.py`
- Add `enable_gravity` to `ArticulationCfg`

## Type of change

## Checklist


embodichain/lab/sim/atomic_actions/affordance.py
```diff
@@ -653,6 +734,133 @@ def _validate_press_position(
...
+            articulation_points - target_center,
+            dim=1,
+        )
+        <= neighborhood_radius
+    )
+    if not bool(neighborhood_mask.any().item()):
+        raise ValueError(f"{field_name} point-cloud neighborhood is empty.")
+    neighborhood_center = articulation_points[neighborhood_mask].mean(dim=0)
+    center_offset = neighborhood_center - target_center
+    axis_index = int(torch.argmax(torch.abs(center_offset)).item())
```
yuecideng: [P2] Resolve the motion axis from articulation kinematics

This always quantizes the neighborhood offset to one Cartesian basis vector. A valid oblique prismatic/revolute axis such as `(1, 1, 0) / sqrt(2)` is therefore returned as `+X`, and independently sampled symmetric clouds can select different signed axes from centroid noise because the rejection tolerance is only `radius * 1e-6`. Slide/Press/Twist then plan a wrong or nondeterministic Cartesian path. Please resolve the active parent joint through `get_parent_joint_chain()`, transform its exact axis into the target-link frame, and use point-cloud geometry only for sign/contact selection (with a confidence check).

embodichain/lab/sim/atomic_actions/affordance.py
```diff
@@ -275,11 +293,41 @@ def __post_init__(self) -> None:
...
 
+    def resolve_from_object_geometry(self, geometry: Mapping[str, Any]) -> None:
+        """Infer the twist axis and its target-local origin from point clouds."""
+        resolved = _infer_articulation_neighborhood_axis(
+            geometry,
+            field_name="TwistAffordance.twist_axis",
+        )
+        if resolved is not None:
+            self.twist_axis, target_points = resolved
+            self.axis_origin = tuple(
```
yuecideng: [P2] Preserve the revolute joint's actual axis origin

The sampled target-link centroid is not guaranteed to lie on the revolute axis (for example, asymmetric or laterally offset knob meshes), so overwriting an explicit origin here makes Twist rotate the grasp around the wrong spatial line. `OpenDoorAffordance.from_articulation()` already shows how to transform the selected joint's `origin_pose` into the target-link frame. Please derive Twist's origin from the active parent joint and keep the centroid for grasp/contact geometry only.

embodichain/lab/task_program/integrations/_configured_services.py
```diff
@@ -356,10 +359,27 @@ def create(
...
             raise TypeError("Articulation must provide get_link_vert_face().")
         vertices, triangles = get_link_vert_face(native_link_name)
+        sample_initial_point_clouds = getattr(
+            articulation,
+            "sample_initial_point_clouds",
+            None,
+        )
+        if not callable(sample_initial_point_clouds):
+            raise TypeError("Articulation must provide sample_initial_point_clouds().")
+        geometry = sample_initial_point_clouds(native_link_name)
```
yuecideng: [P2] Make the explicit translation axis a real compatibility path

`sample_initial_point_clouds()` is called before the optional legacy axis can be used. Existing configured Slide providers that previously supplied a valid axis but have no PK chain, use non-unit `body_scale`, or do not implement this new method now fail during adapter creation; when sampling succeeds, `ObjectSemantics` also overwrites the explicit value. Please skip sampling when `translation_axis` is present, or fall back to that value when sampling is unavailable/unsupported, and cover the factory `create()` path rather than only decoding the field.

yuecideng (changes requested):
I found three medium-severity correctness and compatibility regressions. The bundled Drawer/Microwave paths and CI checks pass, but the public articulation path can infer a wrong or nondeterministic kinematic axis, Twist can rotate about an off-axis mesh centroid, and the documented legacy Slide fallback is not reachable when sampling is unsupported.

| Priority | Area | Impact |
|---|---|---|
| P2 | Axis inference | Oblique axes are quantized to a Cartesian basis; symmetric independently sampled clouds can choose a noise-driven direction. |
| P2 | Twist origin | The mesh centroid can replace a correct revolute-joint origin and produce rotation about the wrong spatial line. |
| P2 | Slide compatibility | Providers with an explicit legacy axis still fail when point-cloud sampling is unavailable or unsupported. |

Requesting changes before merge; details and suggested fixes are attached inline.

embodichain/lab/sim/objects/articulation.py
```diff
@@ -1197,6 +1198,276 @@ def get_link_vert_face(self, link_name: str) -> Tuple[torch.Tensor, torch.Tensor
         verts, faces = self.body_data.link_vert_face[link_name]
         return verts, faces
 
+    def sample_initial_point_clouds(
```
yuecideng: [P2] Keep action-specific geometry sampling out of `Articulation`

This public method combines responsibilities from three layers: `Articulation` topology/FK/raw mesh access, stochastic Open3D surface sampling, and the Atomic Action-specific geometry protocol (`target_link_point_cloud` and `articulation_point_cloud`). Its callers use the result to construct `ObjectSemantics` and resolve affordances, so placing it here makes the simulation-object API depend on PK-chain/unit-scale restrictions and private semantic keys that are not intrinsic properties of an articulation.

Please keep `Articulation` focused on deterministic, domain-neutral facts such as `get_link_vert_face()`, `compute_fk()`, and `get_parent_joint_chain()`. Move this operation to an `atomic_actions/articulation_geometry.py` adapter that accepts an articulation-like provider and returns a typed value. If a class-level API is still needed, expose a deterministic initial mesh snapshot here; perform random sampling and affordance interpretation in the higher layer.

embodichain/lab/sim/objects/articulation.py
```diff
@@ -1197,6 +1198,276 @@ def get_link_vert_face(self, link_name: str) -> Tuple[torch.Tensor, torch.Tensor
...
+        face_areas = 0.5 * torch.linalg.vector_norm(
+            torch.cross(
+                face_vertices[:, 1] - face_vertices[:, 0],
+                face_vertices[:, 2] - face_vertices[:, 0],
+                dim=1,
+            ),
+            dim=1,
+        )
+        valid_faces = face_areas > torch.finfo(vertices.dtype).eps
+        if not bool(valid_faces.any().item()):
```
yuecideng: [P3] Preserve vertex-only links in the merged point cloud

`_validate_point_cloud_mesh()` accepts links with no triangles, and this helper falls back to their vertices only when the entire input has no valid face. In a mixed articulation where the target link has vertices but no (or only degenerate) faces while another link has a valid triangle, `valid_faces.any()` is true and only `triangles[valid_faces]` is passed to Open3D. The target vertices are not referenced by any sampled triangle, so `target_link_point_cloud` is populated while `articulation_point_cloud` silently omits that same link. Downstream target-neighborhood resolution can then be empty or geometrically inconsistent.

Please either apply the fallback per link before merging or reject links without valid triangle surfaces, and add a mixed-case regression test.

yuecideng (commented):
Supplementary review of the new `Articulation` geometry API. The point-cloud functionality is useful, but its current ownership mixes simulation-object facts with stochastic preprocessing and Atomic Action-specific semantics.

| Priority | Area | Recommendation |
|---|---|---|
| P2 | API ownership | Keep `Articulation` deterministic and domain-neutral; move action-specific sampling into a typed geometry adapter. |
| P3 | Mixed mesh handling | Preserve or explicitly reject links that have vertices but no valid triangle faces. |

These comments supplement the existing correctness findings and do not duplicate the earlier joint-axis review.

embodichain/lab/sim/atomic_actions/affordance.py
```diff
@@ -653,6 +747,183 @@ def _validate_press_position(
...
+    neighborhood_mask = (
+        torch.linalg.vector_norm(
+            articulation_points - target_center,
+            dim=1,
+        )
+        <= neighborhood_radius
+    )
+    if not bool(neighborhood_mask.any().item()):
+        raise ValueError(f"{field_name} point-cloud neighborhood is empty.")
+    neighborhood_center = articulation_points[neighborhood_mask].mean(dim=0)
```
yuecideng: **P2 — Do not use target-mesh sampling noise as axis-direction evidence.** The articulation cloud includes the target link, but it and `target_points` are independently sampled by Open3D. If no other link surface falls within `2 × radius`, `center_offset` is only sampling noise; the current tolerance still treats it as directional evidence and can flip the `Slide`, `Press`, or `Twist` axis. Exclude the target link from the sign computation, or preserve the explicit axis / report ambiguity when there is no meaningful non-target neighborhood.

embodichain/lab/task_program/integrations/_configured_services.py
```diff
@@ -356,10 +360,24 @@ def create(
         if not callable(get_link_vert_face):
             raise TypeError("Articulation must provide get_link_vert_face().")
         vertices, triangles = get_link_vert_face(native_link_name)
+        geometry = sample_initial_articulation_geometry(
```
yuecideng: **P2 — Preserve the explicit Slide-axis fallback for scaled articulations.** This always samples before using the compatibility `translation_axis`, but the sampler rejects every non-unit `body_scale`. A pre-existing scaled articulation configuration with an explicit `translation_axis` could create this lowerer; it now fails here and never reaches the fallback. When an explicit axis is provided, keep the legacy path and skip sampling, or implement scale-consistent geometry transformation first.

yuecideng (changes requested):
Request changes: please address the two inline P2 comments before merging. The automatic axis sign can be nondeterministic for target-only neighborhoods, and the new sampler breaks legacy scaled articulation_link_slide configurations even when they provide an explicit axis.

yuecideng (approved):

PR #596, opened 2026-09-07
fix(gym/utils): get_changed_qpos raises on every call (torch.tensor isinstance + wrong helper/keywords)

Anai-Guo:
## Problem

`get_changed_qpos()` in `embodichain/lab/gym/utils/misc.py` cannot complete a
single call — it fails three separate ways:

**1. `torch.tensor` used as a type (line 613)**

```python
if isinstance(qpos_to_change, torch.tensor):
```

`torch.tensor` is a factory function, not a class, so this raises before
anything else runs:

```
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

**2. `replace` mode calls the offset helper with replace keywords (line 632)**

```python
qpos_to_change = get_offset_qpos(
    qpos_to_change,
    replace_value=qpos_change_value,
    replace_joint_list=joint_list_change,
)
```

`get_offset_qpos(qpos_to_change, offset_value, joint_list_offset, degrees=None)`
accepts neither name. The function that does is `get_replaced_qpos(qpos_to_change,
replace_value, joint_list_replace)` — defined just above at line 549 and, before
this change, referenced nowhere in the repository.

**3. `offset` mode misspells the joint-list parameter (line 641)**

```python
offset_joint_list=joint_list_change,   # parameter is named joint_list_offset
```

## Why `get_replaced_qpos` is the intended callee

`get_changed_pose()`, the pose-side twin directly above, already routes each
mode to its own helper:

```python
if change_mode == "replace":
    pose_to_change = get_replaced_pose(pose_to_change, pose_replace_value=..., axis_str_replace=...)
elif change_mode == "offset":
    pose_to_change = get_offset_pose(...)
```

The qpos path just never got wired the same way.

## Verification

`get_replaced_qpos`, `get_offset_qpos` and `get_changed_qpos` were extracted
from `main` with `ast` (no hand-copying) and executed with a stub `log_error`,
before and after the patch:

```
UNPATCHED
  replace: TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
  offset:  TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union

PATCHED
  replace: OK -> [0.  0.9 0.2 0.3]        # qpos_changes=[("replace_1", [0.9])]
  offset:  OK -> [0.  0.27453293 0.2 0.3] # qpos_changes=[("offset_1_degrees", [10.0])], 0.1 + deg2rad(10)
```

Both results match what the helpers' bodies say they should do: `replace` writes
`0.9` into joint 1, `offset` adds `deg2rad(10)` to it.

Changes are confined to the three lines above (+4/-4 in one function); no other
call site of `get_offset_qpos` is touched.

### DIFF
--- a/embodichain/lab/gym/utils/misc.py
+++ b/embodichain/lab/gym/utils/misc.py
@@ -610,7 +610,7 @@ def get_changed_qpos(
     Returns:
         qpos_to_change (np.ndarray): The changed qpos.
     """
-    if isinstance(qpos_to_change, torch.tensor):
+    if isinstance(qpos_to_change, torch.Tensor):
         qpos_to_change = np.asarray(qpos_to_change)
 
     for qpos_change_name, qpos_change_value in qpos_changes:
@@ -629,16 +629,16 @@ def get_changed_qpos(
             )
 
         if change_mode == "replace":
-            qpos_to_change = get_offset_qpos(
+            qpos_to_change = get_replaced_qpos(
                 qpos_to_change,
                 replace_value=qpos_change_value,
-                replace_joint_list=joint_list_change,
+                joint_list_replace=joint_list_change,
             )
         elif change_mode == "offset":
             qpos_to_change = get_offset_qpos(
                 qpos_to_change,
                 offset_value=qpos_change_value,
-                offset_joint_list=joint_list_change,
+                joint_list_offset=joint_list_change,
                 degrees=change_partition[2],
             )
         else:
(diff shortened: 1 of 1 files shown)


yuecideng:
Hi @Anai-Guo , thanks for fixing this issue. Please resolve the review comments from greptile bot.

Anai-Guo:
Thanks @yuecideng! Addressed the greptile P1: the documented `offset_123` (mode_joints) form produces a 2-element `change_partition`, so `change_partition[2]` would `IndexError`. Now guarded as `degrees=change_partition[2] if len(change_partition) > 2 else None` — `get_offset_qpos` already treats `degrees=None` as no degree suffix. Let me know if anything else is needed.

yuecideng (approved):

PR #599, opened 2026-09-08
Add database-driven IK seed selection sampler

Yuan-Xinyi:
## Description

This PR adds `QposSeedSelSampler`, a database-driven joint-seed sampler for iterative IK multi-start, and wires it into `PytorchSolver` behind an opt-in config switch.

Instead of filling the multi-start seed batch with uniform random draws, the sampler retrieves seeds from a lazily built FK database: the joint space is sampled once with a Sobol sequence, and at query time the nearest entries to each target pose (position + Frobenius-rotation metric) are re-ranked by the predicted joint-space correction `||J⁺·Δpose||` so that seeds requiring the smallest correction are tried first.

**Motivation.** Random multi-start wastes most of its seed budget: on the real solver (Franka, 500 reachable targets), random seeding needs `num_samples=16` to reach ≥99% success, while database retrieval reaches it at `num_samples=4` — a ~3.4× end-to-end speedup at equal success rate (584 ms → 169 ms), because well-chosen seeds also converge in ~3× fewer DLS iterations. In a warm-start sweep (perturbed true solution as caller seed, k=4), success at σ=1.0 rises from 87.4% to 99.4%.

**Design.**
- `QposSeedSelSampler` subclasses `QposSeedSampler`: calls without a `target_xpos` fall back to the parent's uniform-random behaviour, and slot 0 always preserves the caller seed, so warm-start waypoint chains (e.g. `MotionGenerator.ik_interp`) are unaffected.
- Opt-in via `PytorchSolverCfg.enable_seed_selection` (+ `seed_db_size`, `seed_rot_scale`); **default off preserves shipped behaviour exactly**.
- The database stores flange poses (no TCP), so runtime `set_tcp()` never invalidates it; `get_ik` queries with its TCP-stripped target. Joint-limit changes trigger an automatic rebuild.
- Cost: ~11 ms / 6 MB one-time build at 20k entries (GPU); <1 ms per `get_ik` query (~0.2% of a solve). Analytic solvers (SRS/OPW/UR) do not consume seeds and are unaffected.

Dependencies: none.

## Type of change

## Screenshots

N/A

## Checklist

- [x] I have made corresponding changes to the documentation (API reference section, `agent_context` ik-solvers topic and MAP keywords)
- [x] Public API changes are reflected in the API docs (`python docs/scripts/check_api_docs.py`: 1826/1826 aligned)
- [x] I have added tests that prove my feature works (`tests/sim/motion/solvers/test_qpos_seed_sel_sampler.py`, 5 tests on the library DexforceW1 robot, including a config-route end-to-end success comparison; existing `test_pytorch_solver.py` passes unchanged)
- [x] Dependencies have been updated, if applicable (no new dependencies)

## Validation

```
pytest tests/sim/motion/solvers/test_qpos_seed_sel_sampler.py  -> 5 passed
pytest tests/sim/motion/solvers/test_pytorch_solver.py         -> 2 passed (regression, default off)
python docs/scripts/check_api_docs.py                          -> 1826/1826
black . / context.py check / agent-context map tests           -> clean
```

### DIFF
--- a/embodichain/lab/sim/motion/solvers/qpos_seed_sel_sampler.py
+++ b/embodichain/lab/sim/motion/solvers/qpos_seed_sel_sampler.py
@@ -1,0 +1,298 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2021-2026 DexForce Technology Co., Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ----------------------------------------------------------------------------
+
+"""Database-driven joint seed selection for iterative IK multi-start.
+
+Instead of filling the multi-start seed batch with uniform random draws,
+:class:`QposSeedSelSampler` retrieves seeds from a precomputed forward-kinematics
+database: the joint space is sampled once with a low-discrepancy sequence, the
+end-effector pose of every sample is stored, and at query time the nearest
+database entries to each target pose are returned as seeds. When a Jacobian
+provider is available, candidates are re-ranked by the predicted joint-space
+step ``||J^+ (target - seed_pose)||`` so that seeds requiring the smallest
+correction are tried first.
+
+The retrieval strategy follows the SELIK solver of the WRS framework
+(Wan Weiwei, 2023, MIT licensed); this is an independent batched PyTorch
+re-implementation adapted to the EmbodiChain seed-sampler contract.
+
+The class is a drop-in extension of :class:`QposSeedSampler`: calls without a
+target pose fall back to the parent's uniform random behaviour, and slot ``0``
+of the returned batch preserves the caller-provided seed so warm-start chains
+(for example sequential waypoint solving) keep their continuity.
+"""
+
+from __future__ import annotations
+
+from typing import Callable
+
+import torch
+
+from embodichain.utils import logger
+
+from .qpos_seed_sampler import QposSeedSampler
+
+__all__ = ["QposSeedSelSampler"]
+
+FkFn = Callable[[torch.Tensor], torch.Tensor]
+"""Batched forward kinematics: ``(N, dof)`` joints to ``(N, 4, 4)`` poses."""
+
+JacobianFn = Callable[[torch.Tensor], torch.Tensor]
+"""Batched geometric Jacobian: ``(N, dof)`` joints to ``(N, 6, dof)``."""
+
+
+def _pose_error_twist(cur: torch.Tensor, tgt: torch.Tensor) -> torch.Tensor:
+    """Compute the ``[dp, rotvec]`` twist between pose batches.
+
+    Args:
+        cur: Current poses with shape ``(N, 4, 4)``.
+        tgt: Target poses with shape ``(N, 4, 4)``.
+
+    Returns:
+        Twist tensor with shape ``(N, 6)``.
+    """
+    dp = tgt[:, :3, 3] - cur[:, :3, 3]
+    rel = tgt[:, :3, :3] @ cur[:, :3, :3].transpose(1, 2)
+    cos = ((rel[:, 0, 0] + rel[:, 1, 1] + rel[:, 2, 2]) - 1.0) * 0.5
+    angle = torch.acos(cos.clamp(-1.0, 1.0))
+    axis = torch.stack(
+        [
+            rel[:, 2, 1] - rel[:, 1, 2],
+            rel[:, 0, 2] - rel[:, 2, 0],
+            rel[:, 1, 0] - rel[:, 0, 1],
+        ],
+        dim=-1,
+    )
+    sin = torch.sin(angle)
+    scale = torch.where(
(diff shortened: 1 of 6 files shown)


embodichain/lab/sim/motion/solvers/qpos_seed_sel_sampler.py
```diff
@@ -0,0 +1,298 @@
...
+            logger.log_error(
+                f"Invalid qpos_seed shape {qpos_seed.shape} for batch_size "
+                f"{batch_size} and dof {self.dof}",
+                ValueError,
+            )
+
+        if self._limits_changed(lower_limits, upper_limits):
+            self._build_database(lower_limits, upper_limits)
+
+        if self._use_caller_seed:
```
matafela, line 316: ### [P2] Preserve the requested seed-batch size when retrieval is capped

When `use_caller_seed=True`, the sampler assumes that `_query()` always returns exactly `num_samples - 1` candidates. That assumption does not hold when either:

- `seed_db_size < num_samples - 1`, or
- `num_samples - 1 > k_max`.

In both cases, `_query()` silently returns fewer candidates because the result is capped by the database size or `k_max`. Unlike the `use_caller_seed=False` branch, the production branch does not pad the result or reject the incompatible configuration.

For example, with `batch_size=3`, `num_samples=5`, and `db_size=2`, the sampler promises a result shaped `(15, dof)` but actually returns `(9, dof)`. `PytorchSolver.get_ik()` still repeats the target poses using the original `num_samples`, so the target batch and initial-configuration batch no longer have matching lengths. This can cause the underlying IK solver to fail and also violates the later reshape assumptions.

Both configurations are currently accepted: `seed_db_size` is only required to be positive, and `num_samples` has no upper bound tied to `k_max`.

Please ensure that `sample()` always returns exactly `batch_size * num_samples` seeds. Possible solutions include padding with valid fallback seeds, increasing the retrieval pool when necessary, or rejecting incompatible values with a clear validation error. Tests should cover both a database smaller than the requested seed count and `num_samples > k_max + 1`.

embodichain/lab/sim/motion/solvers/qpos_seed_sel_sampler.py
```diff
@@ -0,0 +1,298 @@
...
+__all__ = ["QposSeedSelSampler"]
+
+FkFn = Callable[[torch.Tensor], torch.Tensor]
+"""Batched forward kinematics: ``(N, dof)`` joints to ``(N, 4, 4)`` poses."""
+
+JacobianFn = Callable[[torch.Tensor], torch.Tensor]
+"""Batched geometric Jacobian: ``(N, dof)`` joints to ``(N, 6, dof)``."""
+
+
+def _pose_error_twist(cur: torch.Tensor, tgt: torch.Tensor) -> torch.Tensor:
```
matafela, line 58: ### [P2] Handle 180-degree relative rotations correctly in the pose-error calculation


The singularity handling in `_pose_error_twist()` treats every case where `sin(angle)` is close to zero as a small-angle rotation and substitutes a scale factor of `0.5`. However, `sin(angle)` is also close to zero when `angle` is close to π.

At a 180-degree relative rotation, the skew-symmetric terms used to construct `axis` also approach zero. Consequently, the current branch returns an approximately zero rotation vector instead of one whose magnitude is π. A minimal case using the identity rotation as the current pose and a π rotation around the X axis as the target produces a rotation-error norm of `0.0`, whereas the correct norm is approximately `3.14159`.

This directly affects the advertised `||J⁺ · Δpose||` re-ranking step: a candidate with an opposite end-effector orientation can appear to require no rotational correction and may be ranked ahead of a genuinely closer candidate. The IK solver will still perform its own convergence checks, but selecting the wrong multi-start seeds can reduce convergence probability and undermine the feature’s primary purpose.

Please use a numerically stable SO(3) logarithm, such as the repository’s quaternion-to-axis-angle utilities, or implement a dedicated π-neighborhood branch that extracts the rotation axis from the diagonal terms. Focused tests should cover exact π rotations, rotations close to π, and a ranking case demonstrating that a 180-degree-error candidate is not assigned zero correction cost.

matafela (changes requested):

embodichain/lab/sim/motion/solvers/qpos_seed_sel_sampler.py
Yuan-Xinyi: Fixed in 261db3cb. `sample()` now guarantees the `batch_size * num_samples` contract in both branches: a new `_pad_with_random()` fills any retrieval shortfall (database cap or `k_max` cap) with uniform random draws within the limits — the parent sampler's behaviour, so degraded configurations gracefully fall back toward random multi-start instead of desynchronizing from the repeated targets. Padding was chosen over validation because `num_samples` changes per `get_ik` call (the solver syncs it into the persistent sampler), so a constructor-time check could not cover it. Tests cover your exact example (`db_size=2, num_samples=5, batch=3` → `(15, dof)`), `num_samples > k_max + 1`, and the `use_caller_seed=False` branch.

embodichain/lab/sim/motion/solvers/qpos_seed_sel_sampler.py
Yuan-Xinyi: Fixed in 261db3cb. `_pose_error_twist()` now routes the rotation error through the repository's quaternion utilities (`quat_from_matrix` + `axis_angle_from_quat`), which are stable over the whole rotation range — at π the quaternion has `w=0` with a well-defined vector part, so no special-case branch is needed. Verified numerically: exact π returns 3.141593 (previously 0.0), π−1e−3 and small angles are exact to 1e-4. Tests cover exact π, near-π, small/moderate angles, and the ranking case: a 180-degree-flipped candidate now carries cost ≈ π and ranks behind an aligned candidate with a small positional offset.

matafela (approved):

PR #600, opened 2026-09-08
Use joint-range midpoint as default IK seed

Yuan-Xinyi:
## Description

This PR replaces the zero-configuration fallback used when `get_ik` receives no seed with the joint-range midpoint, defined once as `BaseSolver.get_default_qpos_seed()` and consumed by every solver.

**Why the midpoint instead of zero:**

1. **Zero violates joint limits on some arms.** The zero configuration is not a valid joint state on every robot: Franka FR3 excludes it on two joints (`fr3_joint4 ∈ [-3.04, -0.15]`, `fr3_joint6 ∈ [0.54, 4.52]`). A seed outside the limits wastes the multi-start slot, and iterative solvers start from a physically impossible configuration.

2. **Even when feasible, a zero-biased result hurts long-term motion potential.** Nearest-solution selection returns the solution closest to the seed, so a zero seed systematically pulls returned configurations toward wherever zero happens to sit — for asymmetric joint ranges that is close to one of the bounds. A joint parked near its limit has little room left for subsequent motion (tracking, replanning, disturbance recovery). The midpoint is, by construction, the configuration farthest from both bounds: absent any caller preference, it is the neutral choice that preserves maximum future maneuverability.

3. **No negative impact by construction.** The default only takes effect when the caller provides **no** seed. All ten production call sites (action terms, randomization, gizmo, atomic skills, `MotionGenerator.ik_interp`) pass explicit seeds and are byte-for-byte unaffected; the change applies solely to seedless API / tutorial / cold-start use, where no caller preference exists to be overridden.

**Measured on the real `PytorchSolver`** (500 reachable targets, 5 RNG repeats):

| num_samples | zero seed | midpoint seed |
|---|---|---|
| 1 | 17.8% | **71.4%** |
| 4 | 78.6 ±4.8% | **88.8 ±1.1%** |
| 30 | 99.6% | 99.6% (limit margin of returned solutions 0.428 → **0.468 rad**) |

On DexforceW1 (zero feasible) the change is a small consistent gain (+0.2 to +2.6pp), never a regression. The margin improvement at `num_samples=30` persists after success saturates — that is point 2 made visible: same success, but the returned solutions sit measurably farther from the limits.

**Changes per solver.**
- `BaseSolver`: new `get_default_qpos_seed()` — `(lower + upper) / 2`, raises if limits are uninitialized.
- `PytorchSolver`, `NeuralIKSolver`, `SRSSolver` (two paths), `OPWSolver`: zero fallback → default seed.
- `DifferentialSolver`: default seed, expanded to the target batch — the previous `(dof,)` zero seed crashed on batched targets with no seed (latent shape bug).
- `URSolver`: previously had **no** `None` handling and crashed with `AttributeError` at the nearest-solution step; now accepts a missing seed (signature default added).
- `PinocchioSolver`: seedless calls now **reset** `init_qpos` to the default instead of silently reusing the seed left over from the previous call (hidden cross-call state); construction seeds FK with the default, guarded by a model-size check.
- `PinkSolver`: untouched — it already projects the neutral configuration into the limits and is the pattern this PR generalizes.

Dependencies: none.

## Type of change

## Screenshots

N/A

## Checklist

- [x] I have made corresponding changes to the documentation (`agent_context` ik-solvers solver-details)
- [x] Public API changes are reflected in the API docs (`python docs/scripts/check_api_docs.py`: 1825/1825; the new method is covered by the existing `BaseSolver` autoclass entry)
- [x] I have added tests that prove my fix is effective (`tests/sim/motion/solvers/test_default_qpos_seed.py`: infeasibility premise on FR3, midpoint helper contract, None-path ≡ explicit-midpoint equivalence, batched seedless DifferentialSolver)

## Validation

```
pytest tests/sim/motion/solvers/test_default_qpos_seed.py  -> 4 passed
pytest tests/sim/motion/solvers/                           -> full solver suite (regression)
python docs/scripts/check_api_docs.py                      -> 1825/1825
black . / context.py check                                 -> clean
```

### DIFF
--- a/agent_context/topics/ik-solvers/solver-details.md
+++ b/agent_context/topics/ik-solvers/solver-details.md
@@ -93,6 +93,16 @@ and available CUDA backends in seeded and full redundancy-search modes.
 
 ## Seed Sampling and Null-Space Tasks
 
+### Default seed (`BaseSolver.get_default_qpos_seed`)
+
+When `get_ik` receives no seed, every solver falls back to the joint-range
+midpoint from `BaseSolver.get_default_qpos_seed()` — never a zero
+configuration, which violates the limits of some robots (Franka FR3 joints 4
+and 6) and biases nearest-solution selection toward the bounds. Pinocchio
+resets its internal `init_qpos` to this default on seedless calls instead of
+reusing the previous call's seed; UR accepts a missing seed instead of
+crashing. Pink keeps its own limit-projected neutral configuration.
+
 ### `QposSeedSampler` (`qpos_seed_sampler.py`)
 
 Used by iterative solvers (e.g., `PytorchSolver`) to generate joint-seed
--- a/embodichain/lab/sim/motion/solvers/base_solver.py
+++ b/embodichain/lab/sim/motion/solvers/base_solver.py
@@ -276,6 +276,31 @@ class BaseSolver(metaclass=ABCMeta):
         """
         return self.ik_nearest_weight
 
+    def get_default_qpos_seed(self) -> torch.Tensor:
+        """Get the feasibility-safe default IK seed: the joint-range midpoint.
+
+        A zero configuration violates the joint limits of some robots (for
+        example Franka FR3, whose joints 4 and 6 exclude zero), which wastes a
+        multi-start slot, biases nearest-solution selection toward the limits,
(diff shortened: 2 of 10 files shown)


yuecideng (approved):
LGTM

Issue #603, opened 2026-09-09
[Bug Report] UR analytic IK always materializes 512 candidates for single-solution queries


yuecideng:
### Describe the bug

`URSolver.get_ik()` always allocates and computes 512 candidate joint configurations per target pose, even when `return_all_solutions=False` and the caller only needs the nearest valid solution. The 512 candidates are the 8 analytical branches expanded across 2^6 periodic joint-shift combinations. This is functionally useful when all solutions are requested, but it creates substantial avoidable memory traffic and compute overhead for the normal single-solution path.

This is especially visible in workspace Cartesian/plane analysis, where `compute_batch_ik()` can submit thousands of targets in one batch. The workspace analyzer then performs an additional nearest-solution selection in PyTorch after the Warp kernel has already materialized the full candidate tensor.

### Steps to reproduce

1. Run a Cartesian workspace analysis with a UR preset and a sufficiently large sample count:

```bash
embodichain analyze-workspace \
  --robot ur \
  --robot-params '{"robot_type":"ur5"}' \
  --control-part arm \
  --mode cartesian_space \
  --bounds -0.8 0.8 -0.8 0.8 0.0 1.2 \
  --num-samples 10000 \
  --batch-size 1000 \
  --headless
```

2. Profile the call to `URSolver.get_ik()` / `Robot.compute_batch_ik()`.
3. Compare it with `OPWSolver`, which materializes 8 candidates per target for the equivalent analytic IK path.

The relevant implementation currently contains:

```python
# embodichain/lab/sim/motion/solvers/ur_solver.py
N_SOL = 512
all_qpos_wp = wp.zeros(n_sample * N_SOL * DOF, ...)
```

and then computes distances for all 512 candidates before selecting one result when `return_all_solutions` is false.

### Expected behavior

When `return_all_solutions=False`, the UR analytic IK path should avoid materializing the complete candidate tensor where possible. It should either:

- select the nearest valid candidate inside the Warp kernel and return only one solution per target; or
- use a compact, configurable candidate representation and only expand periodic variants when they are required by the joint limits or when `return_all_solutions=True`.

The `return_all_solutions=True` behavior should remain available and preserve all valid periodic representatives.

### System Info

- Commit: `797fe56e` (main)
- OS: Ubuntu Linux (kernel 5.15)
- GPU: NVIDIA RTX PRO 5000 72GB Blackwell
- CUDA: 13.0 (reported by `nvidia-smi`)
[...]

PR #604, opened 2026-09-10
fix ur solver allocate memory

matafela:
# Description

- **Warp kernel:** Added `ur_ik_nearest_kernel` to generate, validate, and select candidates locally, returning one solution per target. Both paths share analytical branch computation.
- **URSolver:** Dispatches by `return_all_solutions`, removing full candidate allocation and PyTorch selection from the single-solution path while preserving limits, periodic representatives, weights, and failure fallback.
- **Tests:** Added 17 CPU/CUDA regression cases covering selection parity, periodic shifts, limits, seeds, ties, and allocation behavior.
- **Benchmark:** Compares the new single-solution, legacy selection, and all-solutions paths at 1k/10k/100k targets, reporting warmed timing, throughput, memory, and accuracy.

Fixes # (issue)

https://github.com/DexForce/EmbodiChain/issues/603

## Type of change

## Checklist

### DIFF
--- a/embodichain/compute/kinematics/_warp/ur.py
+++ b/embodichain/compute/kinematics/_warp/ur.py
@@ -16,9 +16,19 @@
 
 from __future__ import annotations
 
-from typing import Tuple
+from typing import Any, Tuple
 
 import warp as wp
+
+__all__ = [
+    "wp_vec6f",
+    "wp_vec48f",
+    "normalize_to_pi",
+    "URParam",
+    "ur_single_fk",
+    "ur_ik_kernel",
+    "ur_ik_nearest_kernel",
+]
 
 wp_vec6f = wp.types.vector(length=6, dtype=float)
 wp_vec48f = wp.types.vector(length=48, dtype=float)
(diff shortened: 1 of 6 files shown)


embodichain/compute/kinematics/_warp/ur.py
```diff
@@ -498,3 +517,83 @@ def ur_ik_kernel(
...
+                if q < lower_limits[t] - tol or q > upper_limits[t] + tol:
+                    valid = int(0)
+                delta = (qpos_seed.dtype(q) - qpos_seed[i, t]) * joint_weights[t]
+                squared_distance = squared_distance + delta * delta
+
+            if j == 0 and k == 0:
+                best_q = candidate
+                best_valid = valid
+            # Retain the norm (including its rounding) for equal-distance ties.
+            distance = wp.sqrt(squared_distance)
```
yuecideng, line 588: **[P2] Preserve the legacy selection order for equal and near-equal distances**

The scalar sum of squares followed by `wp.sqrt` does not reproduce the rounding of the previous `torch.norm` reduction. Consequently, the strict `<` comparison below does not preserve the promised first-candidate tie behavior: valid seeds near branch bisectors can select a different analytical branch, with joint differences of several radians.

On CPU with PyTorch 2.7.0 and Warp 1.15.0, uniform weights, limits of ±2π, and 3,584 seeds constructed as midpoints between analytical branches, I reproduced 81 selections differing from the legacy path. In 41 cases, the legacy norms were exactly equal but the new kernel selected a later candidate. I also verified that the base commit and the refactored all-solutions kernel produced exactly identical candidate tensors and validity flags, isolating the difference to this selection calculation.

For a concrete float32 example, using the target and seed below selects candidate 256 with the legacy reduction and candidate 384 with the new kernel. Both legacy distances are `2.708728551864624`, and the maximum joint difference is approximately `3.5985` radians. Both solutions are valid; the regression is in selection compatibility.

```python
target = torch.tensor([[
    [0.11713490635156631, 0.8637253642082214, -0.49016112089157104, 0.061536163091659546],
    [-0.34352806210517883, -0.4278513193130493, -0.8360213041305542, -0.12725721299648285],
    [-0.9318088889122009, 0.2663114070892334, 0.24659760296344757, 0.7414405345916748],
    [0.0, 0.0, 0.0, 1.0],
]], dtype=torch.float32)
seed = torch.tensor([[
    0.6716036796569824, -2.2640819549560547, 1.3243775367736816,
    -0.8974207639694214, 0.0, 0.3734279274940491,
]], dtype=torch.float32)
# UR5, identity TCP, uniform weights, all joint limits [-2*pi, 2*pi].
```

Please align the distance reduction/rounding with the legacy selection semantics and add regression coverage for nonzero equal and near-equal distances. The existing zero-weight tie test cannot detect this difference; all 15 added CPU cases passed in this environment.

PR #631, opened 2026-09-15
Visualize per-point manipulability with inspection ellipsoids

Yuan-Xinyi:
## Description

Adds the manipulability visualization requested as item 3 of the review on #614. That PR made `WorkspaceAnalyzer` produce a true per-point Yoshikawa score; this one makes the score readable.

Follow-up to **#614** (now merged); this PR contains only the visualization.

The module keeps a **pure mapping layer** separate from a thin matplotlib renderer, so normalization and alignment are unit-testable without a display:

| Function | Responsibility |
|---|---|
| `align_manipulability_scores` | Tie the compact score vector to the drawn points |
| `normalize_manipulability` | Percentile clipping, optional log10, reports raw **and** clipped ranges |
| `map_manipulability_colors` | Color reachable points, keep unreachable ones distinct |
| `select_inspection_indices` / `inspect_points` | Pick top/bottom/explicit points and gather their diagnostics |
| `translational_manipulability_ellipsoid` | Velocity ellipsoid from the translational Jacobian rows |

Two entry points consume that layer. `workspace_manipulability_visualization.py` writes the Matplotlib figures below, renderer-free. Scene-level coloring belongs to the existing CLI rather than a second runner: `analyze-workspace --vis-type manipulability` draws the colored cloud beside the robot it came from, in the native DexSim viewer or in Viser, with `--manipulability-log-scale` and `--manipulability-percentile` choosing the normalization.

### How each review point is met

- **Alignment.** Joint-space scores map 1:1 onto `workspace_points`; Cartesian and plane scores map 1:1 onto `reachable_points` and are scattered onto `all_points` through `reachability_mask`. The scatter *verifies* `all_points[mask] == reachable_points` and raises rather than drawing a mis-colored figure if that ordering assumption is ever broken.
- **Robust / log normalization with an honest color bar.** Percentile clipping (default 2–98) optionally on `log10`. `clip_range` is converted back into raw `w` units, so the bar, the array, and the footer annotation always agree. The untouched `raw_range` and the saturated-point count are printed next to every figure.
- **Unreachable points stay distinct** — small translucent grey crosses, not dropped and not blended into the ramp.
- **Ellipsoids only where asked.** `inspect_points()` calls `jacobian_fn` **once**, with only the selected rows, and uses `select_jacobian_rows(J, "translational")`. A 6000-point workspace costs a handful of Jacobians, not 6000.
- **Degenerate inputs** are explicit: empty, fully-invalid, and constant score vectors are handled instead of dividing by a zero span.

## Type of change

## Screenshots

Figures 1 to 3 are produced by `scripts/tutorials/sim/workspace_manipulability_visualization.py` on a UR5 and are the figures embedded in the docs page. Figure 4 is the scene view, now produced by `embodichain analyze-workspace --robot ur --vis-type manipulability`.

**1. Cartesian workspace colored by manipulability** — 6000 sampled points: the 981 reachable ones carry the viridis ramp, the 5019 unreachable ones remain grey crosses that outline the sampling box. The footer reports `raw w range [0, 0.1036]` against the percentile-clipped bar span `[0, 0.09719]` and the 20 saturated points, so the clipping is visible rather than silent. Red circles mark the two inspected points.

Cartesian workspace colored by manipulability

**2. Joint-space workspace on a log color scale** — 3541 FK samples with decade ticks on the bar. This is the figure that justifies offering log normalization: the raw range starts at `0` while the clipped span starts at `5.07e-05`, and on a log ramp the near-singular shell stays separable from the dexterous mid-range instead of collapsing into one dark color.

Joint-space workspace on a log manipulability scale

**3. Translational manipulability ellipsoids at the best and worst points** — each panel prints `w`, `cond(J)`, the position, the semi-axes, the anisotropy and the full `qpos`. The best point (`w = 0.104`, `cond = 10.5`) is a near-round ellipsoid; the worst (`w = 0`, `cond = 2.06e9`) is a flat disc whose shortest semi-axis is 0.13 m/s against 0.87.

Translational manipulability ellipsoids at selected points

Worth noting from that last figure: the `w = 0` configuration still has a **three-dimensional** translational ellipsoid. Its rank deficiency lives in the rotational rows, so reading `w = 0` alone would suggest the end-effector cannot move at all, when in fact only the orientation directions are degenerate. That is a concrete argument for showing the ellipsoid next to the scalar, and it is called out in the docs page.

**4. Robot and workspace in one scene** — the same analysis published through the `sim_manager` point-cloud backend, so the cloud can be read against the robot it describes. Analyzer points are arena-frame and already carry the chain-root pose, so the example adds only `sim.arena_offsets[0]` to reach the world frame the scene APIs expect, and never applies the robot base twice.

Robot and colored workspace in the native viewer

## Checklist

- [x] Public API changes are reflected in the API docs
- [x] I have added tests that prove my feature works

## Validation

```
pytest tests/sim/motion/workspace/ tests/compute/test_manipulability.py \
       tests/visualization/test_example_tutorial_coverage.py tests/test_sim_cli.py
                                                                           -> 280 passed
analyze-workspace --robot ur --vis-type manipulability --viser             -> 2363 points, no fallback
python scripts/tutorials/sim/workspace_manipulability_visualization.py     -> 3 figures written
black --check .                                                            -> 1119 files unchanged
python docs/scripts/check_api_docs.py                                      -> 2204/2204 documented
context.py check                                                           -> agent context map: ok
```

74 of those tests are new and cover the mapping and normalization directly: percentile clipping, log scaling, empty / single-point / all-equal inputs, the unreachable mask, index alignment in both result modes, and that the ellipsoid is computed only for the selected subset. Five of them pin the `sim_manager` and `viser` forwarding contracts against a mocked simulation manager, including that a rejected sample cannot shift scores onto a neighbouring point. Three more cover the CLI surface added for the review: the new options, the Viser pass-through alongside the still-falling-back geometry types, and the per-backend point-size unit.

## Review follow-up

Addressing @yuecideng's [P2]: route scene-level manipulability through `analyze-workspace` instead of a second runner (ab0fd8e9).

| Asked | Done |
|---|---|
| Expose `MANIPULABILITY` through `--vis-type`, plus normalization options | `--vis-type manipulability`, `--manipulability-log-scale`, `--manipulability-percentile LOW HIGH` (default `2 98`) |
| Allow it through the analyzer Viser path | The fallback exempts `MANIPULABILITY`; every other non-point-cloud type still falls back |
| Remove the duplicate script or reduce it to an example | `scripts/tutorials/sim/visualize_robot_workspace.py` deleted; project context points at the CLI |
| Keep the renderer-free tutorial | `workspace_manipulability_visualization.py` unchanged |

Normalization lives on `VisualizationConfig`, not only in the CLI, so a programmatic caller configures it the same way. Routing through the CLI also picks up arbitrary presets and assets instead of UR5 only, and caching and cache preview instead of a forced `--no-cache`.

The split surfaced one thing worth naming: the analyzer now selects `point_size` or `viser_point_size` per backend, because the native renderer measures screen pixels and Viser measures scene units, so one value could not be right for both.

Two notes beyond the four points. Metric selection is deliberately untouched: `MetricConfig()` already defaults to `MetricType.ALL`, which includes manipulability, so having the visualization narrow it would only have switched reachability and density off. And offscreen capture went away with the deleted script, since the CLI's `--headless` means "no native window" only; a `--save-image` option can be added if that capability should survive.

## Known limitations

- The color bar and the range annotation are matplotlib-only. The Viser and `sim_manager` backends forward RGB alone, so an in-scene legend would be a separate change, and point size means screen pixels natively but scene units in Viser. Their forwarding contract is unit-tested; the live renderer and browser paths are exercised by the tutorial rather than by CI.
- `ManipulabilityVisualizer` deliberately lets scores override a `colors` array passed by the generic `analyzer.visualize()` path, which hands reachability colors to every visualizer. Documented and covered by a test.

### DIFF
--- a/agent_context/MAP.yaml
+++ b/agent_context/MAP.yaml
@@ -299,7 +299,8 @@ topics:
   title: Robot Workspace
   aliases: [robot workspace, workspace analyzer, reachability cache, 机器人工作空间, 可达空间, 工作空间缓存]
   keywords: [WorkspaceAnalyzer, AnalysisMode, ResultsCache, RobotWorkspace, RobotWorkspaceCfg, workspace_cfg,
-    sample_reachable_pose, analyze-workspace, workspace-cache, results.npz]
+    sample_reachable_pose, analyze-workspace, workspace-cache, results.npz, manipulability_scores,
+    ManipulabilityMetric]
   paths: [topics/robot-workspace/robot-workspace.md]
   source_of_truth:
   - embodichain/lab/sim/motion/workspace/analyzer.py
--- a/agent_context/topics/ik-solvers/ik-solvers.md
+++ b/agent_context/topics/ik-solvers/ik-solvers.md
@@ -132,3 +132,20 @@ solver interfaces. The compute kernels do not import simulation modules.
 `utils/warp/kinematics/*_solver.py` are compatibility aliases.
 Validate kernel import/compilation with `tests/compute/test_imports.py` and
 solver behavior with the corresponding `tests/sim/motion/solvers/` tests.
+
+## Batch adapters and analytic scratch memory
+
+`BaseSolver.get_fk_batch()` and `get_ik_batch()` flatten and restore arbitrary
+leading batch axes in the chain-root frame. Nearest IK returns a boolean success
+mask with the leading shape and qpos with a final `dof` axis; existing concrete
+`get_fk`/`get_ik` signatures remain available. Robot owns local-arena/root frame
+conversion and broadcasts root transforms without materializing per-target copies.
+
+UR/OPW reuse internal candidate buffers through `solvers/_buffers.py`.
+`prepare_buffers(max_batch)` reserves a high-water capacity; allocation is lazy
+and later larger calls grow it. Public results remain independent of subsequent
+calls, including `return_all_solutions=True`. Calls are serialized by a lock and
+CUDA events; Warp kernels use the current Torch stream. Buffer storage is released
+with the solver. UR retains all 512 periodic candidates and the existing nearest
+selection; OPW retains eight candidates. OPW packs live joint limits in one host
+transfer per call, so limit updates do not require cache invalidation.
--- a/agent_context/topics/robot-system/robot-system.md
+++ b/agent_context/topics/robot-system/robot-system.md
@@ -40,6 +40,9 @@ A `Robot` is instantiated with a `RobotCfg` and a list of DexSim `Articulation` 
 - `motion` subpackages load lazily, and workspace analyzer/visualization APIs
   retain a separate lazy export boundary. Do not introduce offline analysis
   imports into the Robot initialization path.
+- `compute_batch_fk`/`compute_batch_ik` broadcast per-environment root transforms
+  and delegate batch shape handling to solver adapters; see
+  [solver batch contracts](../ik-solvers/ik-solvers.md#batch-adapters-and-analytic-scratch-memory).
 - Focused workspace coverage lives under `tests/sim/motion/workspace/`; solver
   tests live under `tests/sim/motion/solvers/`.
 
--- a/agent_context/topics/robot-workspace/analysis-and-cache.md
+++ b/agent_context/topics/robot-workspace/analysis-and-cache.md
@@ -4,7 +4,7 @@
 
 ### Offline analysis, cache, and runtime path
 
-1. `WorkspaceAnalysisConfig` chooses `JOINT_SPACE`, `CARTESIAN_SPACE`, or `PLANE_SAMPLING`,
+1. `WorkspaceAnalyzerConfig` chooses `JOINT_SPACE`, `CARTESIAN_SPACE`, or `PLANE_SAMPLING`,
    plus sampler seed/count/batch size, cache, constraints, visualization, metrics, IK
    seeds/reference pose, control part, and plane settings.
 2. Analysis is designed for one simulation environment; with multiple environments it warns
@@ -14,7 +14,7 @@
 3. Joint-space mode samples within selected joint limits and applies FK; stored
    `workspace_points` and `joint_configurations` contain only valid aligned rows.
 4. Cartesian mode samples requested XYZ bounds, or infers bounds from 1,000 random FK samples
-   plus margin/fallback, then runs seeded IK. It stores all sampled points and reachability
+   plus margin/fallback using one batched FK call, then runs seeded IK. It stores all sampled points and reachability
    outputs plus best joint configurations aligned to reachable points.
 5. Plane mode creates samples on the configured plane and follows the same IK/result alignment path.
 6. `WorkspaceAnalyzer.analyze()` checks `ResultsCache` before computation unless force is set,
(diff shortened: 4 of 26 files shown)


scripts/tutorials/sim/visualize_robot_workspace.py
```diff
@@ -0,0 +1,295 @@
...
+    parser = build_parser()
+    args = parser.parse_args(argv)
+    if args.num_samples < 1:
+        parser.error("--num-samples must be positive")
+    if args.max_steps is not None and args.max_steps < 1:
+        parser.error("--max-steps must be positive")
+    if args.headless and args.backend != "dexsim":
+        parser.error("--headless captures DexSim; use --backend dexsim")
+
+    sim = SimulationManager(
```
yuecideng: [P2] Route manipulability visualization through `analyze-workspace` instead of maintaining a second runner

This script recreates the existing CLI flow—robot construction, `SimulationManager` setup, `WorkspaceAnalyzer` configuration, the display loop, and teardown—while supporting only UR5 and disabling the cache. The established `embodichain analyze-workspace` command already owns this workflow, including arbitrary presets/assets, caching and cache preview, the native viewer, and Viser.

The only scene-level additions here are manipulability coloring/log normalization and offscreen capture. Please expose the new `VisualizationType.MANIPULABILITY` through `analyze-workspace --vis-type` (plus the relevant normalization options) and allow it through the analyzer Viser path, which currently falls back every non-point-cloud type to `point_cloud`. Then remove this script or reduce it to a short usage example. The renderer-free `workspace_manipulability_visualization.py` tutorial should remain: its color bar, raw/clipped range reporting, and selected-point ellipsoid inspection are not provided by the CLI.

yuecideng (changes requested):

scripts/tutorials/sim/visualize_robot_workspace.py
Yuan-Xinyi: Agreed on the premise, and all four points are addressed in ab0fd8e9.

**1. Expose `VisualizationType.MANIPULABILITY` through `analyze-workspace --vis-type` (plus normalization options).**
Done. `--vis-type manipulability` selects the score-colored cloud. Normalization is `--manipulability-log-scale` and `--manipulability-percentile LOW HIGH` (default `2 98`). Both are backed by new `VisualizationConfig` fields rather than CLI-only plumbing, so a programmatic caller configures them the same way. Routing through the CLI also picks up what you listed as already owned there: arbitrary presets and assets instead of UR5 only, and caching and cache preview instead of a forced `--no-cache`.

**2. Allow it through the analyzer Viser path.**
Done. The fallback now exempts `MANIPULABILITY`, which forwards its own colors to the same point-cloud overlay, so Viser renders it directly. Every other non-point-cloud type still falls back, since none has a browser form. One thing the split surfaced: the analyzer now picks `point_size` or `viser_point_size` per backend, because the native renderer measures screen pixels and Viser measures scene units, so a single value was wrong for one of them.

**3. Remove the script or reduce it to a usage example.**
Removed. `scripts/tutorials/sim/visualize_robot_workspace.py` is deleted, and the project context now points at the CLI instead.

**4. Keep the renderer-free `workspace_manipulability_visualization.py`.**
Kept, unchanged. Its color bar, raw versus clipped range reporting and selected-point ellipsoid inspection have no CLI equivalent.

Two notes beyond the four points.

I first also had `--vis-type manipulability` set the metric set, then reverted it. `MetricConfig()` already defaults to `MetricType.ALL`, which includes manipulability, so narrowing would only have switched reachability and density off. Metric selection is untouched.

Offscreen capture went away with the deleted script, and the CLI's `--headless` means "no native window" only. I did not add a replacement, since you asked for the coloring and normalization specifically. Happy to add `--save-image` if that capability should survive.

Verified end to end rather than statically:

```
analyze-workspace --robot ur --mode joint_space --num-samples 4000 \
  --no-cache --headless --viser --vis-type manipulability \
  --manipulability-log-scale
```

publishes 2363 colored points with no fallback warning. Focused tests cover the CLI options, the Viser pass-through alongside the still-falling-back geometry types, and the per-backend point-size unit; reverting the fallback condition fails the middle one. Workspace suite plus the tutorial static checks: 280 passed.

yuecideng (approved):

PR #636, opened 2026-09-16
fix usd articulation import

matafela:
# Description

- Fix usd articulation file import. 
- example: `python scripts/tutorials/sim/create_articulation.py --asset usd`

## Type of change

## Checklist


tests/sim/test_create_articulation_tutorial.py
```diff
@@ -0,0 +1,79 @@
+# ----------------------------------------------------------------------------
```
yuecideng: remove this tutorail test

yuecideng (changes requested):

embodichain/lab/sim/objects/articulation.py
```diff
@@ -2972,15 +2970,15 @@ def compute_jacobian(
...
+        frame_names = self.pk_chain.get_frame_names(exclude_fixed=False)
         if root_link_name is None:
             root_link_name = frame_names[0]  # Default to the first frame
         if end_link_name is None:
             end_link_name = frame_names[-1]  # Default to the last frame
 
         # Create pk_serial_chain
         pk_serial_chain = create_pk_serial_chain(
-            urdf_path=self.cfg.fpath,
+            chain=self.pk_chain,
```
yuecideng, line 3028: [P2] Subset and reorder qpos for the selected serial chain

`compute_jacobian()` now builds a serial chain from the full USD PK tree, but it still passes the full articulation qpos directly to that chain. For the new `DrawerUSD`, the public state contains three sibling joints while the `top_drawer` serial chain contains only `top_slide`; calling `compute_jacobian(drawer.get_qpos(), root_link_name="cabinet", end_link_name="top_drawer")` therefore raises a tensor-size `RuntimeError` in `pytorch-kinematics==0.10.0`. If the Newton state-buffer order differs from the source topology order, equal-width chains can also consume the wrong joint values and expose incorrectly ordered Jacobian columns. Please map the public `joint_names` to `pk_serial_chain.get_joint_parameter_names()`, subset/reorder qpos before calling `jacobian()`, and add a branched-articulation regression test using the three-drawer asset.

yuecideng (approved):

PR #642, opened 2026-09-16
Benchmark seven more atomic skills

Yuan-Xinyi:
## Description

#641 is merged, so this now targets `main` directly and carries only its own
commits.

Extends the suite from five skills to twelve, measured the way #641's standard
defines: an ordered list of stages per skill, each a scene measurement, chained
into one `success_rate`.

```
stage_success_rate[i] = passed(i) / reached(i)      reached(i) = passed(i-1)
success_rate          = passed(last) / total
```

## Results

`--profile coverage --device cpu`, `--repeat 1`, one environment. Full report in
`scripts/benchmark/atomic_action/BENCHMARK_REPORT.md`, per-case tables in
`outputs/benchmarks/`.

```
stage_success_rate[i] = passed(i) / reached(i)      reached(i) = passed(i-1)
success_rate          = passed(last) / measured     measured = total - unsupported
coverage_rate         = measured / total
```

| Skill | Cases | Coverage | Stage chain | `success_rate` |
|---|---|---|---|---|
| `move_joints` | 3 | 100% | reached 100% | **100%** |
| `move_end_effector` | 4 | 100% | reached 100% | **100%** |
| `slide` | 3 | 100% | grasped 100% → slid 100% → released 100% | **100%** |
| `press` | 3 | 100% | contacted 100% → pressed 100% | **100%** |
| `pour` | 3 | 100% | poured 100% | **100%** |
| `twist` | 3 | 100% | grasped 100% → twisted 100% → released 100% | **100%** |
| `axis_align` | 2 | 100% | grasped 100% → aligned 100% → held 100% | **100%** |
| `open_door` | 3 | 100% | grasped 66.67% → opened 100% → released 100% | **66.67%** |
| `pick_up` | 48 | 64.58% | grasped 100% → lifted 74.19% → held 100% | **74.19%** |
| `place` | 48 | 60.42% | released 100% → placed 48.28% → stable 100% | **48.28%** |
| `move_held_object` | 48 | 56.25% | transported 48.15% | **48.15%** |
| `hand_over` | 2 | 100% | grasped → transferred → handed_over 100% → placed 0% | **0%** |

### About 40 % of the grasp cases are ones this embodiment cannot serve

`unsupported_capability` now covers them: the case leaves every denominator and
lowers `coverage_rate`, which is what `BENCHMARK_DESIGN.md` section 5
prescribes and what the standard now states for `success_rate`.

| Skill | Unsupported | In `q1`, near the base | With a `side` approach |
|---|---|---|---|
| `pick_up` | 17 of 48 | 9 of 12 | 13 of 24 |
| `place` | 19 of 48 | 11 of 12 | 14 of 24 |
| `move_held_object` | 21 of 48 | 11 of 12 | 16 of 24 |

Both groups were measured, not assumed. For an object 0.18 m from the base
axis, all three distinct IK branches of the grasp pose exist and driving the arm
to each one stalls, the shoulder stopping 0.44 rad short while every other joint
tracks to 0.01 rad. For a side approach, probing one case at increasing heights,
0 of the sampled grasps are attainable at the object's own height on the ground
and 3 of 3 become attainable once lifted 0.20 m, while the top approach is
attainable at every height — the gripper is hitting the ground.

Both produce a plan that executes into thin air, and the skill's own screen
cannot see either: it accepts a candidate when IK returns a solution, which says
a configuration exists, not that the arm can get to it.
`has_attainable_grasp_candidate()` drives each sampled grasp in physics and
marks the case unsupported only when none of them arrives. **The case set is
untouched** — positions, approaches and object presets are the ones #318
shipped.

Excluding those cases raised `pick_up` from 52.08 % to 74.19 %, `place` from
37.50 % to 48.28 % and `move_held_object` from 27.08 % to 48.15 %, and took
`grasped` and `released` to 100 %: the cases where the skill reported a held
object the scene did not show were all unattainable ones.

### Where the remaining failures are

**`pick_up`**, all eight losses at `lifted`: the hand closes, the arm follows
its plan, the object does not rise 4 cm. Lift over the measured cases has a
median of 0.1563 m, so these are not near misses.

**`place`**, fifteen at `placed`: median XY error 0.0101 m against the 1 cm
tolerance, with a tail to 0.6373 m.

**`move_held_object`**: median 0.0064 m, inside tolerance, with a tail to
0.6478 m — 14 of 26 scored cases land inside 1 cm and the rest are not close.
One further case failed its PickUp precondition and is reported as an
exception.

**`hand_over`**, both cases at `placed` only: 0.1048 m and 0.0680 m against the
derived 0.03 m budget, neither dropped.

**`open_door`**, `open_90` at `grasped`: the hinge never moves.

One recorded failure per skill, captured with the suite's own
`--record_video --record_failed_video --video_case_limit 1`:

PickUp fails to lift the object

Place releases far from the commanded pose

MoveHeldObject leaves the object far from the commanded pose

HandOver delivers outside its budget

### These are single-sample estimates

Everything is `--repeat 1` and grasp sampling is stochastic. `open_door` shows
what that means: the same three cases, the same code, run twice, scored 100 %
and then 66.67 %, because `open_90` grasped the handle on one run and not the
other. Every 100 % above carries that caveat.

### The replay rate is not what produced these numbers

The physical-validation replay moved from 4 to the standard's calibrated 16
physics steps per waypoint. `pick_up` coverage at both rates:

| Steps per waypoint | grasped | lifted | held | `success_rate` |
|---|---|---|---|---|
| 4 (previous) | 87.50% | 59.52% | 100.00% | 52.08% |
| 16 (standard) | 87.50% | 59.52% | 100.00% | 52.08% |

Identical in aggregate; six of 48 cases swap verdicts, three each way.

## What changed since the previous revision of this branch

**Rebased onto `main` with #641 in it.** The branch sat on the pre-rebase stack. Four
of its commits had reached `main` through #620 — their five benchmark files are
now byte-identical to `main`, and replaying them would have **removed** the
`affordance_sampling` support that landed after the branch point. Its own copy
of `BENCHMARK_STANDARD.md` is dropped as superseded, and so is its press/slide
tutorial strategy fix, which is already on `main`: both tutorials request
`strategy="ik_interp"` there today.

**A case the embodiment cannot serve is no longer scored as a skill failure.**
`has_attainable_grasp_candidate()` drives each sampled grasp in physics,
pre-grasp then grasp, and stops at the first one that arrives. When none does,
the case is recorded with `unsupported_capability`, leaves every denominator and
lowers `coverage_rate` instead. Supported cases pay one probe; only an
unsupported one pays for the whole candidate list. `build_stage_leaderboard`
reports `coverage_rate` and `unsupported_cases` alongside the stage rates, and
the standard's section 1 now states the denominator.

**The measurement layer now follows the standard**, which had replaced all three
of the things this branch shipped:

- `StageLadder` and `SKILL_STAGES` replace the four-level ladder
  (`planning_success → motion_valid → execution_success → task_success`).
  Stages are declared per skill for all fifteen; a stage is reached only when
  every earlier one passed, and recording past the first failure is ignored.
- Plan validity and replay tracking are diagnostics. The tracking gate and
  `REPLAY_TRACKING_TOLERANCE_RAD` are gone, and the taxonomy is the nine
  task-level reasons rather than seventeen. `open_door`'s 4x replay rate stays,
  now justified as measurement fidelity rather than as a way past the gate.
- `PRIMITIVE_*` / `TASK_*` tolerances replace `PHYSICAL_PLACE_XY_TOLERANCE_M`
  (0.10) and `PHYSICAL_MOVE_HELD_OBJECT_XYZ_TOLERANCE_M` (0.12).
- `MoveJoints` and `MoveEndEffector` are measured on the robot after a physical
  replay and a terminal settle. From `traj[:, -1]` they reported 0.000000 rad
  and 0.00002 m, which describes the solver; they now measure 0.002996 rad and
  0.0006–0.0015 m.
- `Twist` scores the executed end-effector rotation about the knob axis, with
  the axis probed off the asset. It was failing 0/3 on the knob joint, which
  over-rotates to 2.03–3.14 rad because the gripper wedges against it; that
  value is now a diagnostic and the skill passes 3/3 at ≤ 0.0085 rad.
- `HandOver`'s delivery budget is the derived `3 x TASK_POSITION_TOLERANCE_M`
  instead of a borrowed 0.12 m.
- Release and settle stages are measured rather than assumed: `hand_is_released`
  compares the hand against the skill's own open and closed commands, and
  `dropped_below_support` applies the cross-cutting drop rule to the lowest
  point of a replay.
- `replay_and_track_channels` samples several scalars in one replay, so a skill
  scored on one signal can report another; a missing actuation segment now
  fails the case as `invalid_case` instead of silently scoring the trajectory's
  last waypoint, which is the rebound the helper exists to avoid.

### Two defects on the aggregate `--in_process` path

Found while validating this branch, both fixed here.

`_make_child_args()` built the child namespace from a hand-kept list of the
five original benchmarks' arguments, so every action added here raised
`AttributeError` on its own case selection — `door_cases`, `press_cases`,
`slide_cases`, `twist_cases`, `align_cases`, `pour_cases`, `hand_over_cases` —
the smoke profile included. The namespace now starts from the selected module's
own `add_benchmark_args()` defaults, with the aggregate run's shared settings
applied over them.

With that fixed, `--action all --in_process` reached `hand_over` and failed with
`Robot DualUr5HandOver not found`: the simulator is a singleton keyed by
instance id, and leaving the previous benchmark's scene standing makes the next
one a second instance whose planner resolves against the first benchmark's
robots. `HandOver`'s own teardown helper moved to `common.release_simulation()`
and the dispatcher calls it after each in-process benchmark.

`--action all --profile smoke --device cpu --in_process` now runs all twelve and
writes twelve reports. Neither defect touched the measurements above: every
coverage run was a direct module invocation, and both failures produced no
report rather than a wrong number.

## Type of change

## Screenshots

N/A — reports are Markdown tables written to `outputs/benchmarks/`.

## Checklist

- [x] Public API changes are reflected in the API docs (none; benchmark scripts)
- [x] I have added tests that prove my feature works

## Validation

`tests/benchmark/atomic_action`, 41 tests.
`test_measurement_standard.py` carries 13 of them. Three of
them read `BENCHMARK_STANDARD.md` and check the shipped stage table, failure
taxonomy and tolerance constants against the document, so code and standard
cannot drift apart. The rest cover the stage bookkeeping the benchmarks depend
on: a stage after a failure is never reached, a closing stage fails the case,
out-of-order recording and non-taxonomy reasons raise, an unreached stage is
`N/A` rather than zero, diagnostics change no outcome, stage rates divide by the
cases that reached the stage, and the two shared scene rules.

`test_aggregate_dispatch.py` covers the `--in_process` path: parametrized over
every registered action, it compares the child namespace against the `args.*`
attributes each benchmark module actually reads, checks each action keeps its
own case-selection defaults, and asserts the simulation is released between
benchmarks. Against the namespace those commits replace, it fails for all seven
added actions.

```
python -m pytest tests/benchmark/atomic_action -q     # 41 passed
black --check scripts/benchmark/atomic_action tests/benchmark/atomic_action
```

All twelve benchmarks were run at `--profile coverage`; the commands are in the
report's reproducing section.

## Open

1. **`move_held_object` has a half-metre tail.** Median position error is inside
   the 1 cm tolerance, but 12 of 26 scored cases are not close, the worst at
   0.6478 m. First thing this suite should be pointed at.
2. **`pick_up` loses every one of its cases at `lifted`** — the hand closes on a
   reachable grasp and the object does not come up. `place`'s fifteen `placed`
   failures are the same kind of question one stage later.
3. **`press`'s criterion does not discriminate commanded depth.** Commanding 10,
   20 or 30 mm all drive the button to 6.00 mm of its 6 mm stroke.
4. **Single-sample estimates.** Everything is `--repeat 1`; `open_door` scored
   100 % and then 66.67 % on two runs of the same code.
5. **The unsupported probe is a benchmark-side measurement**, not a skill
   change. It says the arm cannot reach a grasp; it does not tell PickUp to
   choose differently. Whether the skill should screen candidates by
   executability is a separate question, and a bigger one: nothing in these
   runs shows it picking a bad candidate when a good one existed.
6. **Three skills still have no benchmark module:** `push_object`,
   `coordinated_pickment`, `coordinated_placement`.


matafela (dismissed):

scripts/benchmark/atomic_action/run_benchmark.py
```diff
@@ -43,6 +43,13 @@
     "pick_up": "scripts.benchmark.atomic_action.pickup_benchmark",
     "move_held_object": "scripts.benchmark.atomic_action.move_held_object_benchmark",
     "place": "scripts.benchmark.atomic_action.place_benchmark",
+    "open_door": "scripts.benchmark.atomic_action.open_door_benchmark",
```
matafela, line 46: [P2] Initialize action-specific defaults for in-process dispatch
The newly registered actions are also exposed through --in_process, but _make_child_args() does not populate their required case-selection attributes, such as door_cases, press_cases, and align_cases. Each new benchmark reads its attribute directly, causing an AttributeError; all seven new actions are affected, including the smoke profile. Please initialize the child namespace through the selected module’s add_benchmark_args() defaults before applying shared overrides, and cover this dispatch path with a focused regression check.

matafela (changes requested):

scripts/benchmark/atomic_action/run_benchmark.py
Yuan-Xinyi: Confirmed and fixed in 2b3304af, with a second defect on the same path fixed in 540da09d.

**Reproduced.** All seven newly registered actions, the smoke profile included:

```
python -m scripts.benchmark.atomic_action.run_benchmark \
    --action press --profile smoke --device cpu --in_process
AttributeError: 'Namespace' object has no attribute 'press_cases'
```

`open_door` → `door_cases`, `press` → `press_cases`, `slide` → `slide_cases`, `twist` → `twist_cases`, `axis_align` → `align_cases`, `pour` → `pour_cases`, `hand_over` → `hand_over_cases`. The five original benchmarks were unaffected because the hand-kept namespace in `_make_child_args()` happened to list exactly their arguments.

**Fix, as you suggested.** The child namespace now starts from the selected module's own `add_benchmark_args()` defaults, and the aggregate run's shared settings are applied over them. Registering a new action no longer requires editing a list in the dispatcher. Two details worth noting:

- `object_types`, `position_cases` and `approach_cases` are forwarded only when the aggregate run was actually given them, so a benchmark otherwise keeps its own default selection.
- A shared setting is applied only where the module declares it, so a skill that samples no grasps is not handed `n_sample` or `force_reannotate`.

**Regression check** in `tests/benchmark/atomic_action/test_aggregate_dispatch.py`, parametrized over every registered action: it extracts the `args.*` attributes each benchmark module actually reads and asserts the child namespace carries them, plus checks that each action keeps its own case-selection defaults and that shared settings override correctly. I rebuilt the pre-fix namespace and ran the same assertions against it: it fails for all seven actions, so the test does guard this regression.

**A second defect, which the first one was hiding.** With the arguments fixed, `--action all --in_process` reaches `hand_over` and fails there:

```
ValueError: Robot DualUr5HandOver not found
  base_planner.py:220, SimulationManager.get_instance(...).get_robot(...)
```

The simulator is a singleton keyed by instance id and a planner resolves its robot through that instance. Each benchmark builds its own scene, so leaving the previous one standing makes the next benchmark a second instance, whose planner then looks for its robot among the first benchmark's robots. The subprocess path never sees this because every action gets a fresh process. `HandOver` already tore its scene down between its own cases; that helper moved to `common.release_simulation()` and the dispatcher now calls it after each in-process benchmark. A test asserts the `run → release → run → release` ordering.

**End to end.** `--action all --profile smoke --device cpu --in_process` now runs all twelve benchmarks in one process and writes twelve reports, exit 0. `tests/benchmark/atomic_action` is 41 tests, all passing.

For the record, this did not affect any measurement in the report: every coverage run was a direct module invocation (`python -m scripts.benchmark.atomic_action.<skill>_benchmark`), which never touches the dispatcher, and the failure mode was an immediate `AttributeError` that produced no report rather than a wrong number.

matafela (approved):

scripts/benchmark/atomic_action/move_held_object_benchmark.py
```diff
@@ -412,25 +458,23 @@ def _on_validation_step(waypoint_index: int) -> None:
...
-        )
-        success = physical_move_success
-        if success:
-            failure_reason = ""
-        elif not is_success:
-            failure_reason = "planning_failed"
+        ladder = StageLadder(stages=SKILL_STAGES["move_held_object"])
+        if not is_success:
+            ladder.fail("transported", "planner_reported_failure")
         elif not still_held:
```
yuecideng, line 464: [P1] **Measure retention from the physical replay**

`still_held` is derived from `result.projected_context`, which is a hypothetical state produced during side-effect-free compilation before this replay runs. `MoveHeldObject` preserves that symbolic held relation whenever planning succeeds, so this guard remains true even if the gripper physically loses the object. If the dropped object happens to settle within the pose tolerances below, the `transported` stage can pass. Determine retention/drop from replay observations at the action boundary and keep projected state diagnostic-only.

yuecideng (commented):
Additional inline findings that are not already covered by the existing review threads.

scripts/benchmark/atomic_action/axis_align_benchmark.py
```diff
@@ -0,0 +1,545 @@
...
+        # The tracked scalar is already the axis angle error to the target.
+        ladder.record(
+            "aligned",
+            trace.measured_position <= TASK_ROTATION_TOLERANCE_RAD,
+            "task_goal_miss",
+        )
+        dropped = dropped_below_support(height.min_position, height.initial_position)
+        ladder.record(
+            "held",
+            (not dropped) and trace.settled_position <= TASK_ROTATION_TOLERANCE_RAD,
```
yuecideng, line 277: [P2] **Verify that the object is still held**

This `held` stage checks only that the object never fell more than the drop margin and that its settled axis remains aligned. An object released onto the same support surface can satisfy both conditions while no longer being grasped, so AxisAlign can report full success after losing the object. The standard defines this stage as `still held`; add physical grasp-retention evidence (for example contact or stable hand/object relative pose) at the held boundary.

scripts/benchmark/atomic_action/pour_benchmark.py
```diff
@@ -0,0 +1,578 @@
...
+    else:
+        ladder.max_tracking_error_rad = trace.max_tracking_error_rad
+        # Pour rotates to the poured pose and then returns to the pose it
+        # started from, all inside one "pour" segment, so the end of the
+        # segment is back at zero rotation by construction. The achieved pour
+        # is the peak signed rotation reached during the segment.
+        rotation_error = abs(
+            abs(trace.peak_signed_displacement) - abs(case.rotate_angle_rad)
+        )
+        ladder.record(
```
yuecideng, line 295: [P2] **Apply the cross-cutting drop rule to Pour**

The replay samples only the object's rotation, so a case that reaches the commanded angle and then drops the object during the return or terminal hold still records `poured=True`. `BENCHMARK_STANDARD.md` requires any skill holding an object to fail the current stage with `object_dropped` when the replay crosses the drop margin. Track object height in the same replay and include that result in this verdict.

scripts/benchmark/atomic_action/place_benchmark.py
```diff
@@ -398,25 +442,31 @@ def _on_validation_step(waypoint_index: int) -> None:
...
-            failure_reason = "physical_pick_failed"
-        elif place_xy_error_m is None:
-            failure_reason = "physical_replay_missing"
+        ladder = StageLadder(stages=SKILL_STAGES["place"])
+        if not is_success:
+            ladder.fail("released", "planner_reported_failure")
         else:
-            failure_reason = "place_target_miss"
+            ladder.record("released", released, "release_failure")
+            ladder.record(
```
yuecideng, line 450: [P2] **Score `placed` at the actuation boundary**

`place_xy_error_m` is computed from `final_obj_position`, which `replay_trajectory_for_physical_validation()` returns only after its terminal hold, even though `segment_end_pose` was captured during replay. An object can therefore miss the target when Place stops acting, roll into tolerance during settle, and pass `placed`; the reverse is attributed to the wrong stage. Compute this stage from the object pose at the placing/release boundary and reserve the settled pose for `stable`.

PR #643, opened 2026-09-16
Use manipulability to diversify trajectory augmentation

### LINKED ISSUE #634: [Proposal] Use manipulability to diversify trajectory augmentation for downstream policy training
### Proposal

Use manipulability to guide trajectory augmentation in `embodichain/lab/sim/motion/expansion/`, with the goal of generating a richer distribution of task-valid trajectories that improves downstream model generalization and robustness.

### Motivation

Manipulability describes how a robot's local motion capability changes along a trajectory. It can guide sampling across different valid postures, paths, and motion bottlenecks, including lower-manipulability configurations that still allow successful execution. Covering these variations helps avoid concentrating training data around a small set of preferred configurations.

### Additional context

Reuse the existing manipulability computations to characterize trajectories and guide candidate generation and selection. Preserve geometric diversity and independent checks for motion feasibility and task success. Evaluate the benefit through distribution coverage and downstream performance on unseen configurations and perturbed execution conditions.

### Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

Yuan-Xinyi:
# Description

Expert trajectory collection currently samples joint residuals uniformly around one reference path, so the accepted rollouts tend to cluster around the same few comfortable arm postures. This change lets manipulability guide both how candidates are proposed and which ones are kept, so a run covers a wider range of postures, including tighter ones that still complete the task.

The work stays inside `embodichain/lab/sim/motion/expansion/` and reuses the existing Yoshikawa helpers in `embodichain.compute.kinematics`; no new numerical algorithm is introduced.

**New module `expansion/manipulability.py`**

- `describe_manipulability()` scores caller-supplied Jacobians and reports the per-sample values together with the lowest value inside each trajectory phase. Scoring runs in float64: the determinant of a float32 Jacobian underflows to exactly zero at small scales and would report a healthy posture as singular.
- `ManipulabilityBands` groups scores into ordered ranges expressed as ratios against a per-case reference value, usually the reference trajectory's lowest score. One set of boundaries therefore transfers across robots and tasks.
- `manipulability_guided_residual()` draws several `joint_residual` proposals from one local generator, scores each, and keeps the first one that lands in the requested range, falling back to the nearest reachable range. Proposals that leave the joint limits are skipped; if every proposal fails, the last error is raised rather than swallowed.

**Operator permission check**

`manipulability_guided_residual()` stays in `manipulability.py` rather than moving next to the primitive operators, because it composes `joint_residual()` with scoring and the two modules would otherwise import each other. It still rejects a template that does not authorize `joint_residual` before entering the proposal loop, so a permission error is never caught by the retry handler and reported as a rejected sample. That check previously reached into `operators._allowed_phases`, the only cross-module use of a private name in the package; it is now the public `operators.allowed_phases()`, so an operator implemented outside that module has a supported way to validate template permissions. No behaviour changes.

**Selection side**

`CoverageIndex` gains an optional per-range quota. Once a range is full, further rollouts in it are rejected even when their measured geometry is new, so one well-conditioned posture cannot consume the whole collection budget while tighter postures still have room. `GenerationSession` classifies each rollout from the `manipulability` observation measured during execution, never from a planned estimate, and reports per-range commit counts in `snapshot()`.

**Compatibility**

Everything is opt-in through `augmentation.factors.manipulability`, which is disabled by default. Existing jobs decode, validate, and run exactly as before; `register_case()` accepts a reference value only when the factor is enabled, and `CoverageIndex` keeps its current behaviour when no quota is configured. Jacobians are supplied by the caller, so the package still imports no robot, solver, or simulation code. Manipulability only ranks postures: path, dynamic, and task validation are unchanged and remain the sole source of feasibility evidence.

## Figures

`scripts/tutorials/sim/planner/manipulability_augmentation_plot.py` overlays twelve guided variants on one UR5 reference, four steered at each band, and writes two views of that same set in one run so they cannot drift apart. The robot is an asset-backed kinematic chain and solver, so the figure needs no simulation manager, renderer, physics backend or display, and it is byte-identical across runs.

Manipulability-guided joint residuals on one UR5 reference

Three things are readable from it. Every variant meets the reference exactly at both phase endpoints, which is the contract `joint_residual` enforces and the reason contacts and grasps stay untouched. The interiors separate by target band rather than scattering uniformly: nine of twelve reached the band they asked for, and the variant bottlenecks span `[0.0301, 0.0534]` around a reference bottleneck of `0.0392`. The three that missed are drawn dashed and keep the color of the band they actually reached, since reporting the nearest reachable band is the defined fallback rather than a failure.

The reference deliberately passes through an interior waypoint near the wrist singularity. A reference whose bottleneck sits on an endpoint cannot have that bottleneck moved by any residual, so every variant would report the reference score and the figure would show nothing.

The joint paths show that targeting moved something. The companion figure shows what it moved it to: manipulability measured along those same paths, with the band edges drawn as ratios of the reference bottleneck.

Manipulability measured along the same variants

Each curve carries a marker at its own bottleneck, which is the single value its band membership is read from. All thirteen curves meet at both ends because the endpoints are shared, and the dip is the interior waypoint. The markers separate across the `0.8x` and `1.2x` edges rather than clustering at the reference, which is band targeting working. Scores are the profile each variant was selected with, not a fresh evaluation, so the curves are exactly the evidence the operator ranked.

**Relationship to #651**

This is the lower layer of a stack. It contributes the scoring, the reference-normalized bands and the guided sampler, but it deliberately adds no orchestration: in this package the host drives `propose()` and calls the operator itself, so `manipulability_guided_residual()` has no in-package caller here by design.

#651 introduces that orchestration (`expand_trajectory_variants()`), already takes per-sample Jacobians, and enumerates joint-path methods from a closed `SPATIAL_METHODS` registry. Wiring this operator into that registry belongs there, on top of this change, so each layer stays one logical change. The two touch different fields of `factors` and different members of `CoverageIndex`, so they compose without redesign.

Not included: a coverage or downstream-training comparison. Measuring the benefit needs real scene collection plus policy training, which is outside this algorithm package; it is a reasonable follow-up once a host integration adopts the option.

No new dependencies.

**Merged with `main`**

The agent-context streamlining change rewrote both context pages this branch had edited, so their old structure was gone. `simulation-system.md` is taken from `main` unchanged: its per-subsystem inventory table no longer exists, and augmentation now routes to the motion topic through `main`'s own tables plus this branch's `MAP.yaml` keywords. The `expansion/manipulability.py` contract was re-applied onto `main`'s "Trajectory augmentation boundary" section, keeping every constraint it stated. No source file conflicted.

Fixes #634

## Type of change

## Validation

- Focused augmentation and manipulability tests: `python -m pytest tests/sim/motion/expansion tests/compute/test_manipulability.py` — **215 passed**. 29 of them are new cases in `tests/sim/motion/expansion/test_manipulability.py`, plus added cases for the range quota in `test_coverage.py`, the new configuration fields in `test_cfg.py`, and the session's use of measured evidence in `test_session.py`.
- `black .` with Black 26.3.1: 1118 files unchanged.
- `python docs/scripts/check_api_docs.py`: **2187/2187 exports documented**.
- `python scripts/tutorials/sim/planner/manipulability_augmentation_plot.py`: both figures written, byte-identical on re-run.
- `context.py check`: agent context map ok.
- The full test suite and any simulation-backed tests were not run; the change is confined to one package with no runtime or backend behaviour outside it.

## Checklist

- [x] I have made corresponding changes to the documentation (expansion API reference, sim overview, motion-planning project context).
- [x] I have added tests that prove the feature works.
- [x] Dependencies have been reviewed; no updates are required.


embodichain/lab/sim/motion/expansion/session.py
```diff
@@ -181,34 +207,44 @@ def register_case(
...
             previous,
             previous_limits,
             previous_names,
+            previous_reference,
         ) in self._cases.items():
             if case_id == case.scene_case_id and (
                 replace(previous, initial_state_id=case.initial_state_id) != case
                 or not torch.equal(previous_limits, limits)
                 or previous_names != names
+                or previous_reference != manipulability_reference
```
matafela: [P2] Keep manipulability references specific to each initial state

With manipulability enabled, registering two initial states in the same scene fails when their reference trajectories have different bottleneck scores. I reproduced this with references of 1.0 and 0.1: the second registration raises ValueError, despite identical scene and robot conditions. Reference manipulability depends on the trajectory and should be allowed to vary between initial states.
Store and resolve the normalizer using (scene_case_id, initial_state_id), while keeping coverage quotas shared by scene. Removing the equality check alone is insufficient because _bands and _measured_band() currently use only scene_case_id. Add a regression test covering registration and classification with different references.

tests/sim/motion/expansion/test_manipulability.py
```diff
@@ -0,0 +1,342 @@
...
+    assert rotational == pytest.approx(0.125)
+    assert describe_manipulability(jacobian, rows="translational").bottleneck == (
+        pytest.approx(1.0)
+    )
+
+
+def test_low_precision_jacobians_are_scored_without_determinant_underflow() -> None:
+    # float32 det() of a 1e-4-scaled Jacobian underflows to exactly zero and
+    # would report a healthy posture as singular.
+    jacobians = diagonal_jacobians(torch.full((2,), 1e-4)).to(torch.float32)
```
matafela: [P3] Make the underflow regression test reject the broken implementation

The current 3×3 Jacobian scaled by 1e-4 produces a determinant around 1e-24, which does not underflow in float32. Additionally, pytest.approx(1e-12) uses a default absolute tolerance of 1e-12, so an incorrect zero result passes. I confirmed that the existing test still passes after removing float64 promotion in memory.
Use a 6×6 diagonal Jacobian scaled by 1e-4, which actually produces zero with float32 scoring, and assert against pytest.approx(1e-24, rel=1e-6, abs=0). This makes the test protect the numerical behavior it claims to verify.

embodichain/lab/sim/motion/expansion/session.py
Yuan-Xinyi: Fixed in 13098351.

A reference manipulability is the bottleneck of one reference trajectory, not a scene condition, so it is now stored and resolved by `(scene_case_id, initial_state_id)`:

- `_bands` is keyed by `(scene_case_id, initial_state_id)`, and `_measured_band()` resolves the normalizer with the same key. Constructing `ManipulabilityBands` also moved out of the `scene_case_id not in self._coverage` branch — otherwise the second initial state's reference would have been dropped silently, which is the point you raised about removing the equality check alone.
- The cross-initial-state check in `register_case()` no longer compares `manipulability_reference`. Scene signature, joint limits and joint names must still match.
- Re-registering the same initial state with a changed reference still fails with `A registered case's fixed conditions cannot change.`
- Coverage quotas stay shared by scene case: one `CoverageIndex` with its `target_per_band` per scene case.

Regression test `test_initial_states_may_hold_different_manipulability_references` in `tests/sim/motion/expansion/test_session.py` covers both registration and classification: two initial states register references of 1.0 and 0.1, the same measured score of 0.3 lands in band 0 for the first and band 2 for the second, the snapshot reports `manipulability_bands["case"] == {0: 1, 2: 1}`, and a changed reference for one initial state is still rejected. Against the previous implementation the test raises `ValueError` at `session.py:233`.

`python -m pytest tests/sim/motion/expansion tests/compute/test_manipulability.py` — 215 passed.

tests/sim/motion/expansion/test_manipulability.py
Yuan-Xinyi: Fixed in 13098351. Your reading is confirmed by measurement:

| Jacobian | dtype | det(J Jᵀ) | score |
|---|---|---|---|
| 3×3 @ 1e-4 | float32 | 9.9999972e-25 | 9.999999e-13 |
| 3×3 @ 1e-4 | float64 | 1e-24 | 1e-12 |
| 6×6 @ 1e-4 | float32 | **0.0** | **0.0** |
| 6×6 @ 1e-4 | float64 | 1e-48 | 1e-24 |

float32 represents the 3×3 determinant of ~1e-24 without any underflow, and `pytest.approx(1e-12)` carries a default absolute tolerance of exactly 1e-12, so a zero result passed as well. Neither half of the test held.

The test now scores a 6×6 Jacobian, whose determinant of 1e-48 is below the smallest float32 subnormal (~1.4e-45) and does collapse to zero, and asserts `pytest.approx(1e-24, rel=1e-6, abs=0)` so no absolute tolerance can absorb a wrong answer.

Verified that the test rejects the broken implementation: with the float64 promotion removed from `describe_manipulability()` it fails at `test_manipulability.py:111`, and passes once the promotion is restored.

yuecideng (approved):

PR #637, opened 2026-09-16
Add geometry-constrained parallel affordance expansion

matafela:
## Description
Add `--n_affordance_expand` N to generate and execute affordance variations across N parallel environments. Variations follow task geometry, using valid grasp candidates, declared symmetries, or accepted motion ranges.

#### Main changes

- Introduce AffordanceSamplingContext and AffordancePoseCandidates for reproducible sampling, explicit candidate validity, and diversity-based selection. Preserve the nominal branch and report candidate reuse when distinct alternatives are exhausted.
- Connect candidate selection to action planning: (B, K, 4, 4) candidate poses become one selected pose per environment, then action-specific waypoints and (B, T, D) joint trajectories. affordance_sampling.py handles selection; action planners retain ownership of IK and trajectory generation.
- Preserve PickUp’s feasibility checks and each selected grasp’s object-to-EEF transform for subsequent transport and placement.
- Support constrained variations for grasping, axis alignment, pressing, door opening, sliding, twisting, interaction points, and assembly symmetries. Missing optional motion ranges preserve exact task targets.
- Count max_episodes as accepted per-environment episodes and consistently commit partial final batches across dataset, trajectory, and camera recording.
- Add a temporary Default/CUDA drawer-handle material refresh at startup to address first-environment grasp slip. The refresh restores the original friction immediately and leaves mass, inertia, and joint state unchanged. 
- This change also updates the shared UR5 gripper grasp target from 0.024 to 0.040, the shared trajectory sample count from 40 to 100, and drawer hand interpolation from 12 to 18. Shared component changes affect all deployments referencing those components.

#### Example commands

conda activate embodichain2
```
embodichain run-env \
  --gym_config embodichain_tasks/configs/tasks/manipulation/repeated_pick_place/task.ur5.yaml \
  --device cuda --headless --seed 42 \
  --max_episodes 180 --n_affordance_expand 9
```
```
embodichain run-env \
  --gym_config embodichain_tasks/configs/tasks/manipulation/open_drawer/task.ur5.yaml \
  --device cuda --headless --seed 42 \
  --max_episodes 180 --n_affordance_expand 9
```
Nine branches mean nine total environments. With all rows accepted, 20 episodes require three batches (9 + 9 + 2), while 180 episodes require twenty batches.

Append --record_trajectory --trajectory_save_dir ./outputs/affordance_trajectories to save joint trajectories.

#### Documentation

[...]

PR #651, opened 2026-09-17
Add trajectory variants between fixed waypoints

Yuan-Xinyi:
## Description

Affordance sampling (#644) varies **where** the robot makes contact. This PR covers the other half: once a plan's waypoints are settled, produce several different ways to execute them, so downstream imitation learning and reinforcement-learning post-training see more than one solution per task instead of many copies of one.

The two compose. A sampled affordance produces a new set of waypoints, and variant expansion then produces several ways of executing that set.

This also gives `motion/expansion/` its first caller. The package already shipped candidate contracts, coverage bookkeeping and `GenerationSession`, but nothing in the repository used them, and only two of its seven declared augmentation factors were implemented.

It follows #644's structure: the algorithms live at the simulation layer, a tutorial is the demonstration host, and **Task Program, Gym lifecycle, dataset persistence and `GenerationSession` collection bookkeeping are explicitly deferred**. No `embodichain_tasks/` file is touched.

Refs #644

#### Main changes

**Operators** (`expansion/operators.py`)

- `via_points` routes an allowed free phase through interior knots, interpolated with clamped cubic Hermite segments. `joint_residual` adds one fixed-shape bump per phase; with two or more knots this changes the shape of the path, not only its amplitude. The knots' signed magnitudes vary along **one** sampled joint direction: sampling every joint of every knot independently decorrelates the joints, and the tool path that forward kinematics produces then wanders — measured at 1.51x the reference arc length on the Place reference, against 1.11x once the knots share a direction.
- `nullspace_residual` projects a residual onto the null space of task Jacobians supplied by the caller, changing arm posture while holding the declared task rows. A fully constrained task raises rather than silently returning the reference.
- `retime` gains a bounded within-phase `profile` (`uniform`, `ease_in`, `ease_out`) that redistributes time inside a phase without changing the path or the phase's total duration. The `uniform` path is arithmetically unchanged from the existing operator.
- `perturb_approach_direction` places standoff poses on a cone around a nominal approach direction while leaving the contact transform exact.
- **Bug fix surfaced by running this:** joint-limit rejection previously scanned the whole trajectory, including joints no operator touches. The tutorial's reference holds `gripper_finger2_joint_1` at −8e−6, just under its `0.0` lower bound, which rejected **23 of 24** proposals while reporting "Sampled residual violates joint limits". The check now covers only the joints an operator actually moved. This also affected the pre-existing `joint_residual`.

Every qpos operator uses an envelope that is zero, with zero derivative, at both endpoints of the phase it modifies, and `contact` and `hold` phases are never touched. Annotated waypoints, contact windows and dwell durations stay bit-identical to the reference.

**Variant generation** (`expansion/variants.py`, new)

- **`spatial.method` names one or more joint-path methods.** It was a single-choice string, so "enumerate every joint-path operator" could not be expressed at all. It now accepts a sequence, each requested method becomes its own variant, and a bare string is still accepted where a sequence is expected, so the existing spelling keeps working.
- **One entry point, one number.** `expand_trajectory_variants(template, case=..., count=8, joint_limits=..., control_dt=...)` is the whole interface, with every implemented method enabled by default. Everything else is an advanced option: `default_variant_factors` to read or adjust the policy, `cfg=` for exact control, `plan_trajectory_variants` / `apply_trajectory_variant` to separate enumeration from application, and the optional motion-limit and attempt-budget arguments. The tutorial mirrors the split: `--trajectory_variants` is the only variant flag an ordinary run needs, and the remaining eight sit in an `advanced variant options` group so `--help` shows the simple path first. With no advanced flag set the tutorial calls the helpers with no configuration at all, so the documented simple path is the one actually exercised.
- **Configuration is optional.** `expand_trajectory_variants(template, case=..., count=6, ...)` is the whole API for "give me six different ways to run this". Omitting `cfg` resolves `default_variant_factors`, which enables every implemented factor the supplied inputs support at the magnitudes measured below, and drops the null-space factor when no Jacobians are given rather than proposing variants that would be rejected. An explicit `cfg` is never overridden, and an explicitly enabled factor that cannot produce a variant now raises instead of disappearing. That covers `ik` without `task_jacobians`, and `approach`, whose Cartesian standoff poses have to be replanned through IK first: the configuration schema still accepts the factor so an upstream planning stage can declare it, but the qpos entry points refuse it rather than silently producing no approach variation. The resolved settings come back on `TrajectoryVariantSet.cfg`, so an auto-configured run stays reproducible.
- `plan_trajectory_variants` lists the enabled factor combinations. Ordinal zero is always the unmodified reference; later ordinals cycle through the enabled joint-path operators, then the duration scales, then the time warps.
- `apply_trajectory_variant` applies one combination, running at most one joint-path operator so a null-space projection is never stacked on an already displaced path.
- `expand_trajectory_variants` collects variants for one fixed scene, rejecting proposals that an operator refuses, that fail sampled motion limits, or whose measured geometry and timing duplicate an accepted variant. Each rejection is counted under a key naming its reason, so a caller can retune the offending setting instead of guessing.
- `CoverageIndex.family_of` exposes the resolved geometry family, so `geometry_family_id` reflects measured grouping rather than a fresh digest.

**Configuration** (`expansion/cfg.py`)

`ik` and `approach` were `_DisabledFactorCfg` stubs that raised when enabled; they are now real settings. `spatial` gains `via_count` and `timing` gains `profiles`. `contact`, `contact_timing` and `recovery` remain unimplemented and are still rejected when enabled.

**Tutorial hook** (`scripts/tutorials/atomic_action/place.py`, `tutorial_utils.py`)

No new tutorial. Following #644, the shared logic lives in `tutorial_utils.py` and the existing Place tutorial grows a small hook (+40/-3): `--trajectory_variants` maps one variant per simulation row, the same shape `--affordance_branches` uses. Row zero always replays the plan unchanged.

The tutorial declares only which of the action's own named `TrajectorySegment` ranges may move — gripper-close and release are contact phases, and only motion at or after the lift is retimed so the shared `clear_dynamics()` step index stays aligned. Everything else comes from the default policy. Advanced overrides sit in their own `advanced variant options` argument group.

`--variant_plot_dir <dir>` writes two figures:

- **Joint trajectories.** Every variant's arm joints against time. Curves separate between waypoints and rejoin inside the shaded contact phases. The title carries the largest contact-phase deviation of any variant from the reference — `0.00e+00 rad`, so contact windows are preserved exactly rather than approximately.
- **Tool paths.** Every variant's tool-centre path in the arena, beside the arm drawn at its starting configuration and the cube drawn to scale on the ground. Stars mark the annotated waypoints, and every variant draws its own, so a waypoint an operator had moved would show up as a scattered cluster instead of a single star; the title reports how far apart they land, which is `0.0e+00 m`, together with the lowest tool centre (2.8 cm) so ground clearance is stated rather than implied by a viewing angle.

A junction handler removes the zero-duration sample each concatenated action repeats at a join, after verifying it is a duplicate; a join that actually moves the robot is refused rather than discarded.

#### What this does and does not guarantee

- `nullspace_residual` holds the declared task rows to **first order only**. Phase endpoints stay exact because the envelope vanishes there, but interior samples drift with linearization error and need forward-kinematics verification by the host.
- The Jacobian's reference frame and column order are the **caller's** responsibility. `BaseSolver.get_jacobian` returns a base-frame Jacobian, so dropping its angular-z row removes rotation about base z, not about the tool axis. Those coincide for a top-down grasp, which is why the tutorial keeps rows `(0, 1, 2, 3, 4)`. Declaring `controlled_joint_indices` in solver order keeps the columns aligned without any permutation.
- Deduplication compares measured joint geometry and elapsed phase time. It is a similarity measure, not a validity proof. Nothing here establishes collision freedom, dynamic feasibility or task success.
- The tutorial's lift check is an observation, not a task-success contract.

## Figures

Both are committed under `docs/source/_static/trajectory_variants/` and embedded in the overview page, so they survive the branch. Produced by the command below with a hundred variants.

**Arm joints of 100 variants against time.** Curves separate between waypoints and rejoin inside the shaded contact phases. The title carries the largest contact-phase deviation from the reference: `0.00e+00 rad`. The wide green band on `joint6` is `nullspace_residual` — for a grasp taken from directly above, wrist roll is the freedom the declared task rows leave open.

Arm joints of 100 trajectory variants

**Tool paths of the same 100 variants.** Stars are the annotated waypoints; every variant draws its own, so they would scatter if an operator had moved one. They spread by `0.0e+00 m`. The arm is drawn at its starting configuration and the cube to scale on the ground.

Tool paths of 100 trajectory variants

Measured over those hundred: all accepted, no rejected proposals, a hundred distinct geometry families, median tool-path length 1.09x the reference and 1.38x at worst, and three variants whose lowest tool centre dips under 2 cm against a median of 2.8 cm. Sampling more variants reaches further into those tails, which argues for a host-side clearance or collision check rather than trusting the generator alone.

## Calling it

Configuration is optional; every implemented factor is enabled by default.

```python
from embodichain.lab.sim.motion.expansion import expand_trajectory_variants

result = expand_trajectory_variants(
    template,                                  # reference with annotated phases
    case=case,                                 # scene and initial-state identity
    count=100,                                 # how many variants to collect
    joint_limits=robot.get_qpos_limits()[0],
    control_dt=0.01,
    task_jacobians=jacobians,                  # optional; unlocks posture variants
)
for variant, row in zip(result.variants, range(len(result.variants))):
    print(variant.spatial_operator, variant.duration_scale, variant.timing_profile)
    positions = result.candidates.positions[row][: result.candidates.valid_length[row]]
```

The tutorial does exactly this behind one flag:

```
python scripts/tutorials/atomic_action/place.py --headless --num_envs 1 --device cuda \
  --trajectory_variants 100 --variant_plot_dir outputs/variants
```

## Type of change

- [x] Documentation update

## Validation

Automated checks:

| Check | Result |
|---|---|
| `black --check ./` | 1119 files pass |
| `python docs/scripts/check_api_docs.py` | 2197/2197 exports documented |
| `context.py check` and routing | ok |
| `pytest tests/sim/motion/expansion` | 195 passed |
| `py_compile` on the touched and neighbouring tutorials | passed |
| `pytest tests/test_agent_context_*.py tests/docs` | 299 passed, 1 failed |
| `python -m py_compile` on the tutorial | passed |

The one `tests/docs` failure is `test_project_myst_configuration_remains_valid`, which raises `ModuleNotFoundError: No module named 'myst_parser'`. Sphinx and MyST are not installed in this development environment, so that test and the Sphinx docs build could not run here; both are unrelated to this change and covered by the CI `build` job.

**The tutorial was launched on a GPU (RTX 4090) and its results are below**, unlike the affordance sampling tutorials in #644, which that PR could not launch:

```
python scripts/tutorials/atomic_action/place.py --num_envs 6 --device cuda --variant_plot_dir outputs/variant_plots

Trajectory variant 0: operator=none,               scale=1.0,  uniform, samples=239
Trajectory variant 1: operator=joint_residual,     scale=1.0,  uniform, samples=239
Trajectory variant 2: operator=via_points,         scale=1.0,  uniform, samples=239
Trajectory variant 3: operator=nullspace_residual, scale=1.0,  uniform, samples=239
Trajectory variant 4: operator=joint_residual,     scale=1.25, uniform, samples=277
Trajectory variant 5: operator=via_points,         scale=1.25, uniform, samples=277

max contact-phase deviation from the reference: 0.00e+00 rad
```

Six distinct geometry families covering all three joint-path operators, no rejected proposals, and exact preservation of every contact phase. The two figures are written to `--variant_plot_dir`; they are not committed, since `outputs/` is gitignored. This run exercised the real UR5 solver Jacobian through `nullspace_residual`, which unit tests alone could not.

### Measured tuning

`joint_offset_scale` is a fraction of each joint's **declared** range, and this arm declares ±2π per joint, so small fractions are large in absolute terms:

| `joint_offset_scale` | Rows lifting the cube |
|---|---|
| 0.005 | 6 of 6 |
| 0.01 (tutorial default) | 6 of 6 |
| 0.02 | 5 of 6 |
| 0.03 | 5 of 6 |

The usable window is narrow on a task this constrained, and it interacts with deduplication: at the shared default of 0.01 for both `joint_offset_scale` and `coverage.joint_dedup_normalized_tol`, the `via_points` variant collapses into the reference's geometry family. The tutorial therefore defaults `--dedup_tolerance` to 0.004. Both numbers are documented rather than silently tuned.

### Default configuration

The schema's own defaults leave every factor off, set `joint_offset_scale` to 0.05, `via_count` to 1 and `target_per_cell` to 1. Enabling factors one at a time from those values reproduces every problem measured above: a silent single-variant result, offsets past the point where the grasp survives, `via_points` degenerating to a single bump, and timing variants deduplicated away. Those schema defaults are left untouched, since an all-off configuration is a meaningful explicit state; the working values live in `default_variant_factors` and apply when a caller does not supply a configuration at all.

### Naming

An earlier revision called these "trajectory modes". That word was invented for this change; the expansion package already says "variant" (`CoverageIndex` documents "timing variants per geometry") and #637 says "variations". The API, configuration, module, docs and tests use `variant` so the change reuses the repository's existing vocabulary instead of adding a parallel term.

## Checklist

- [x] I have added production-contract tests that prove the feature works
- [x] Dependencies have been updated, if applicable (no dependency changes required)


embodichain/lab/sim/motion/expansion/cfg.py
```diff
@@ -122,26 +141,89 @@ def __post_init__(self) -> None:
...
+        if self.via_count > _MAX_VIA_POINTS:
+            raise ValueError(f"spatial.via_count must be at most {_MAX_VIA_POINTS}")
+
+
+@configclass
+class _IkCfg:
+    enabled: bool = False
+    method: str = "nullspace_residual"
+    normalized_scale: float = 0.05
+    task_rows: tuple[int, ...] = (0, 1, 2, 3, 4)
```
yuecideng, line 183: [P1] **Apply `ik.task_rows` or remove the public knob** — This field is validated and documented as a configuration option, but no code in the variant application path reads it; the full `task_jacobians` tensor is passed straight to `nullspace_residual`. A caller selecting a subset of spatial rows therefore gets the same projection as the default, with no warning. Select these rows at one clearly owned boundary (and test a full six-row input with a custom subset), or remove the field and document that callers must pre-reduce the Jacobian.

embodichain/lab/sim/motion/expansion/variants.py
```diff
@@ -0,0 +1,631 @@
...
+            candidate = apply_trajectory_variant(
+                template,
+                variant,
+                cfg=cfg,
+                joint_limits=joint_limits,
+                control_dt=control_dt,
+                task_jacobians=task_jacobians,
+                generator=generator,
+            )
+        except ValueError as error:
```
yuecideng: [P1] **Do not swallow malformed inputs as proposal rejections** — `apply_trajectory_variant` raises `ValueError` for structural or caller errors as well as for an operator rejecting a sampled proposal (for example invalid Jacobian or joint-limit shapes, or an invalid control period). Catching every `ValueError` here converts those programming/configuration errors into rejected-count bookkeeping and can return a partial result after the nominal variant has already succeeded. Prevalidate inputs before the attempt loop or introduce a dedicated rejection exception and catch only that type.

embodichain/lab/sim/motion/expansion/operators.py
```diff
@@ -134,46 +297,203 @@ def joint_residual(
...
+            dtype=torch.float64,
+        ).to(q.device)
+        offsets = (2 * values - 1) * scale * spans
+        window = projector[phase.start_index : phase.stop_index].matmul(offsets)
+        largest = max(largest, float(window.abs().max()))
+        envelope = _endpoint_envelope(length, window)
+        q[phase.start_index : phase.stop_index, indices] += (
+            envelope[:, None] * window
+        ).to(q.dtype)
+    if largest <= 0:
```
yuecideng: [P2] **Use a numerical rank test instead of `largest <= 0`** — For a generic full-rank, non-orthogonal Jacobian, `I - pinv(J) @ J` contains floating-point residuals even though the null space is mathematically zero, so `largest` will usually be positive. The current check therefore only reliably rejects special cases such as an identity Jacobian and may emit a near-unchanged variant when no redundancy exists. Use the singular values/rank with the existing `rank_tolerance`, or compare the projector effect against a scale-aware tolerance, and add a non-diagonal full-rank regression test.

embodichain/lab/sim/motion/expansion/cfg.py
Yuan-Xinyi: Fixed in e727f51a by applying the rows rather than removing the field, since row selection is genuinely an ik-factor decision.

`apply_trajectory_variant` now calls `select_jacobian_rows(task_jacobians, cfg.factors.ik.task_rows)` before projecting, so that is the one owned boundary. The contract flipped accordingly: `task_jacobians` is the full spatial Jacobian and callers pass every row their task could constrain instead of pre-reducing. The tutorial no longer carries its own `task_rows` parameter, which was the duplicate decision.

Regression test as requested — `test_configured_task_rows_select_from_a_full_spatial_jacobian` feeds a six-row Jacobian on six joints and asserts that keeping all six raises for lack of redundancy while dropping one succeeds, with the residual confined to the single unconstrained joint. That fails if the knob is ignored.

embodichain/lab/sim/motion/expansion/variants.py
Yuan-Xinyi: Fixed in e727f51a with the dedicated exception.

`ProposalRejected(ValueError)` is raised only for draw-dependent outcomes, which after auditing the operators is exactly one: a sampled residual leaving the joint limits. Everything else — unusable phase lengths, an absent or malformed Jacobian, an unaligned command clock, an unknown operator — stays a plain `ValueError` and propagates. `expand_trajectory_variants` catches `ProposalRejected` only.

The partial-result path you describe was real: the nominal variant applies no operator, so a malformed Jacobian only surfaced on ordinal 1 and was then counted as a rejection, leaving a one-row result that looked like an unlucky run. `test_a_malformed_jacobian_is_not_counted_as_a_rejected_proposal` pins that it now raises, and `test_a_limit_violation_is_a_rejection_not_a_malformed_input` pins that the two cases carry different types.

Subclassing `ValueError` keeps existing `except ValueError` callers working.

embodichain/lab/sim/motion/expansion/operators.py
Yuan-Xinyi: Confirmed and fixed in e727f51a. I measured it before changing anything:

```
                        largest       rank   old guard
identity 6x6            0.000e+00      6/6   rejects
random full-rank 6x6    8.135e-13      6/6   PASSES
ill-conditioned 6x6     7.464e-14      6/6   PASSES
```

So the guard only ever caught the identity case — which is precisely what the old test used, making it a test that could not fail for the reason it claimed.

The check is now `torch.linalg.matrix_rank(jacobians, rtol=rank_tolerance) >= len(indices)`, evaluated once before the phase loop at the same tolerance the pseudoinverse uses, and the `largest` accumulator is gone.

`test_nullspace_residual_refuses_any_fully_constrained_task` is parametrized over identity, random and ill-conditioned full-rank inputs, asserting rank 6 on each first so the fixture cannot silently degenerate. `test_nullspace_residual_accepts_a_rank_deficient_task` covers the other side.

yuecideng (changes requested):
## Review summary

I reviewed the current head `e727f51a9fca95131f7c4363570c2af37f11bfee`. The focused expansion tests pass locally (202 tests), but the following issues should be addressed before merge.

### P1 — Default planning and application do not share the resolved configuration

[`variants.py:250`](https://github.com/DexForce/EmbodiChain/blob/e727f51a9fca95131f7c4363570c2af37f11bfee/embodichain/lab/sim/motion/expansion/variants.py#L250)

`plan_trajectory_variants(count)` defaults to `default_variant_factors()`, which enables `nullspace_residual`. A caller that then applies the returned variants through `apply_trajectory_variant(...)` with its default configuration and no Jacobians reaches the null-space variant and gets `ValueError: nullspace_residual requires task_jacobians`. The default plan→apply workflow therefore cannot complete. Please pass the resolved configuration through the workflow, or disable redundancy when the required Jacobians are not available.

### P1 — Direct application silently ignores an enabled approach factor

[`variants.py:386`](https://github.com/DexForce/EmbodiChain/blob/e727f51a9fca95131f7c4363570c2af37f11bfee/embodichain/lab/sim/motion/expansion/variants.py#L386)

`apply_trajectory_variant` dispatches spatial qpos operators and retiming, but never checks `cfg.factors.approach.enabled`. Passing an approach-enabled configuration to this public entry point can return an unchanged qpos trajectory even though the caller requested approach variation. This silently drops a configured factor. Please fail fast here, consistently with `plan_trajectory_variants` and `expand_trajectory_variants`, or route approach variation through the required Cartesian replanning path.

### P2 — Generic scalar-to-tuple coercion bypasses timing schema validation

[`cfg.py:104`](https://github.com/DexForce/EmbodiChain/blob/e727f51a9fca95131f7c4363570c2af37f11bfee/embodichain/lab/sim/motion/expansion/cfg.py#L104)

The decoder converts every `tuple[str, ...]` scalar into a one-element tuple. Consequently, malformed input such as `timing.profiles: "uniform"` is accepted and normalized, while direct construction rejects the same scalar and the schema otherwise requires a sequence. Please limit this compatibility conversion to `spatial.method`.

### P2 — `TIMING_PROFILES` is missing from the defining module's public exports

[`operators.py:36`](https://github.com/DexForce/EmbodiChain/blob/e727f51a9fca95131f7c4363570c2af37f11bfee/embodichain/lab/sim/motion/expansion/operators.py#L36)

`TIMING_PROFILES` is documented and re-exported by the package, but it is absent from `embodichain.lab.sim.motion.expansion.operators.__all__`. Star imports and module-level API discovery therefore omit a documented public symbol. Please add it to `__all__`.

### P3 — Crowded plots lose timing provenance

[`tutorial_utils.py:844`](https://github.com/DexForce/EmbodiChain/blob/e727f51a9fca95131f7c4363570c2af37f11bfee/scripts/tutorials/atomic_action/tutorial_utils.py#L844)

When more than twelve variants are plotted, labels are grouped only by `spatial_operator`. Timing-only variants, or variants sharing one spatial operator but using different duration scales/profiles, become indistinguishable black curves with no legend entry describing their timing. Please preserve timing provenance in the grouping or annotate it in the plot.

### Validation

- Local overlay of the PR files: `pytest tests/sim/motion/expansion -q` — 202 passed.
- PR checks: lint, build, and documentation checks passed; the selected test job and test gate are currently failing because the GPU lane exits with `-11` in `tests/sim/motion/solvers/test_opw_solver.py`.

Because the two P1 issues affect public default/configured workflows, I am requesting changes before merge.

Yuan-Xinyi:
All five are fixed in `8c0e2e5d`. Focused suite: 263 passed (202 before, plus the four regression tests below and one existing assertion updated).

### P1 — Default planning and application do not share the resolved configuration

Confirmed: `plan_trajectory_variants(6)` followed by `apply_trajectory_variant(...)` raised on the first `nullspace_residual` ordinal.

The root cause is that neither entry point can see whether the caller has Jacobians, so neither can safely resolve the permissive default. Both now resolve conservatively instead:

- `plan_trajectory_variants(count)` with no `cfg` resolves `default_variant_factors(redundancy=False)`, so every variant it offers is applicable with nothing but a reference trajectory.
- `apply_trajectory_variant(...)` with no `cfg` resolves `default_variant_factors(redundancy=task_jacobians is not None)`, so the same variant runs the posture operator when the Jacobians are there and is unreachable when they are not.

`expand_trajectory_variants` is untouched: it resolves the configuration once against the inputs it actually received and passes that object down, and it still returns it on `result.cfg`. A caller that wants posture variants out of the split workflow passes `default_variant_factors(redundancy=True)` to both calls; that path is now tested too.

Three regression tests: the unconfigured plan→apply pair completes for every ordinal and offers no `nullspace_residual`; the explicit redundant policy does offer it and applies cleanly with Jacobians; and one planned posture variant applies with Jacobians and raises without them.

### P1 — Direct application silently ignores an enabled approach factor

Fixed. The guard that `plan_trajectory_variants` already had is now a shared `_reject_cartesian_factors(cfg)` called from both entry points, so an approach-enabled configuration raises in `apply_trajectory_variant` instead of returning an unchanged qpos trajectory. Routing the factor through Cartesian replanning stays out of scope here — the standoff pose has to go back through IK, which is a host-side capability this package does not own; `perturb_approach_direction` remains the operator a host calls before planning.

### P2 — Generic scalar-to-tuple coercion bypasses timing schema validation

Fixed, narrowed exactly as suggested. `_decode` now consults a `frozenset` of `(class, field)` pairs holding only `(_SpatialCfg, "method")`, rather than keying off the annotation shape. `timing.profiles: "uniform"` and `ik.task_rows: 0` now fail with `must be a sequence`, matching direct construction; `spatial.method: "via_points"` still decodes, which is the backward-compatible spelling this shim existed for. Parametrized test added over both previously-wrapped fields.

### P2 — `TIMING_PROFILES` is missing from the defining module's public exports

Added to `operators.__all__`.

### P3 — Crowded plots lose timing provenance

Fixed by using the second visual channel rather than by dropping the grouping. Past the threshold, colour still names the spatial operator and the dash pattern now names the timing signature (`profile` + duration scale), with a second legend titled "timing" beside the geometry legend. Dash patterns for signatures past the built-in table are stretched copies rather than repeats, so two signatures can never collide — the first attempt did collide at six signatures, `uniform x1.25` reusing solid from `ease_in x1`, which is visible if you diff the figures against the previous revision.

Both committed figures are regenerated from the same 100-variant run, and the ≤12-variant path is unchanged apart from label word order.

### Note on the failing GPU lane

`tests/sim/motion/solvers/test_opw_solver.py` exiting `-11` is unrelated to this branch; it reproduces on `main`.

Yuan-Xinyi:
Correction to the last line of my previous comment: I said the OPW solver failure "reproduces on `main`", which I had not verified and should not have asserted. What I can actually support is narrower — this branch adds no diff under `embodichain/lab/sim/motion/solvers` or `tests/sim/motion/solvers`, and `pytest tests/sim/motion/solvers/test_opw_solver.py` passes in this worktree (4 passed, 4 skipped), so the GPU lane's `-11` is not produced by these changes. Whether it also fails on `main`'s GPU lane I have not checked.

yuecideng (approved):

PR #668, opened 2026-09-22
Update microwave asset

matafela:
# Description
- Update microwave asset for OpenDoor, Twist and Press atomic action example/benchmark.
- Larger trajectory interpolation number in twist and press action,  make sure that robot end-effector reaches the target correctly.

## Type of change

## Checklist

- [x] I reviewed affected documentation and agent context, updated it where needed, or explained why no update was needed.


Yuan-Xinyi:
Verified the new asset locally: md5 matches, `handle_link` / `button_link` / `knob_joint` all exist, 4 movable joints so `init_qpos` arity still fits, and all three tutorials run headless with `success=True`. Three things to fix.

**1. The rename did not reach the prose.** These still name links and joints that no longer exist in `Microwave/microwave.urdf`:

- `scripts/benchmark/motion_generation/BENCHMARK_DESIGN.md:28-29` still says Press and Twist use `button_cap`/`start_button_press` and `cap_1`/`power_knob_rotation`. This is the design doc for the three suites this PR edits, so it now contradicts the YAML.
- `docs/source/overview/sim/atomic_actions/builtin_actions.md:845` still says the OpenDoor example configures the microwave's `door_handle` link.
- `scripts/tutorials/atomic_action/open_door.py:121,124` docstring still says it resolves from `door_handle` through `door_to_door_handle_fixed`. In the new URDF those are `handle_link` and `door_handle_fixed_joint`, while the constant right above it is already correct.

**2. Dead line in `twist.py:67`.** The old asset path is left commented out, while `open_door.py` and `press.py` deleted theirs.

**3. The sample-count bumps are unexplained and not forced by the planner.** The PR body says only "Update microwave asset", but it also changes:

| where | param | before | after |
|---|---|---|---|
| main + pose_batch | press `sample_count` | 140 | 256 |
| main + pose_batch | twist `sample_count` | 140 | 256 |
| main + pose_batch | twist `hand_interp_steps` | 12 | 20 |
| randomized | press `sample_count` | 80 | 256 |
| randomized | press `hand_interp_steps` | 8 | 12 |

`Press._motion_segment_lengths` needs `sample_count >= hand_interp_steps + 8` (20) and `Twist._motion_segment_lengths` needs `sample_count >= 2 * hand_interp_steps + 8` (48), so 140 and 80 were already well above the budget. `sample_count` is the trajectory waypoint count and therefore the IK call count, so the randomized suite goes 3.2x in wall clock and previously recorded numbers stop being comparable. Could you add the rationale to the PR body, or split this out?

**Two notes, not blocking.**

The asset's DOF semantics changed and the body does not mention it: movable joint 2 went from revolute `timer_knob_rotation` (±pi) to a limitless continuous `turntable_joint`, the knob narrowed from ±pi to ±2.356 rad, and the door hinge from 2.007 to 1.920 rad. Nothing breaks today since `init_qpos` is all zeros and the configured -0.785 rad twist and 60 deg default open angle both still fit, but it is worth stating.

No test loads the new URDF. The changed tests assert YAML values or run mock planners, so asset compatibility has no CI coverage. The new parametrizations in `test_actions.py` are a good addition though, and they cost about 2 seconds.

Checks run here: 154 passed in `tests/sim/atomic_actions/test_actions.py`, 58 in the two benchmark config tests, 12 in `tests/data`, `black --check` clean on all 7 files. The benchmark's physical replay assertions were not run.

Yuan-Xinyi (commented):
Please check whether the attached comments are reasonable.

matafela:
> Please check whether the attached comments are reasonable.

I resolved the issue in comments.

Yuan-Xinyi (approved):

PR #680, opened 2026-09-23
Define joint and EEF action contracts for datasets

### LINKED ISSUE #601: [Proposal] Define joint and EEF action representations for expert and LeRobot datasets
### Proposal

Define an explicit action-representation contract for expert datasets, including LeRobot exports: when should a dataset store joint targets, end-effector (EEF) pose targets, or both?

Keep the existing joint-target representation as the default for joint-controlled demonstrations. Evaluate opt-in Cartesian action representations for policies deployed through an EEF controller, while distinguishing requested Cartesian targets from measured EEF poses and executed joint commands.

This issue records a design proposal, not a request to make EEF pose a mandatory action field or immediately introduce a new trajectory abstraction.

### Motivation

Related: #595 introduces timed-trajectory velocity targets and explicit expert position/velocity action schemas.

A demonstration may pass through several representations:

```text
Cartesian goal -> IK / motion planning -> timed joint targets -> controller -> observed motion
```

These values serve different purposes. A Cartesian goal supplied once per segment is not automatically a per-control-step action, and an EEF pose computed from observed joints is not the commanded target. Mixing them produces ambiguous training labels and replay behavior.

We need to support:

- Joint-space policies trained on qpos or qpos + qvel targets.
- Cartesian policies trained on absolute EEF targets or pose increments, with an explicit deployment controller.
- Auxiliary EEF observations and target labels for analysis, supervision, and trajectory visualization.
- Future cross-embodiment experiments, without assuming Cartesian actions alone guarantee transfer across robots.

#### Proposed direction

1. **Match the training action to the policy's deployment interface.** Joint-controlled datasets should retain joint targets by default. Cartesian actions should be opt-in and require a defined Cartesian controller/IK contract. Record both representations where useful, but explicitly identify which is the training action and which is an auxiliary target or executed command.
2. **Separate observations from commands.** Store measured/FK-derived EEF poses as observations. Store requested EEF targets separately when available. Do not silently infer a commanded EEF target from the observed pose.
3. **Use an explicit schema.** Evaluate how to express primary actions and auxiliary numeric fields in the supported LeRobot version, including compatibility with existing loaders and training pipelines. Avoid silently concatenating joint and Cartesian targets into one ambiguous action vector.
4. **Preserve temporal alignment.** Define the relationship between observation at step t, command at step t, and the subsequent transition. Specify how segment-level goals and resampled targets align with control-step records.
5. **Keep dataset ownership separate from planning results.** Assess whether existing PlanResult fields and dataset-boundary adapters are sufficient before adding or expanding trajectory container types. A planner's EEF path is not automatically a dataset action sequence.

#### Schema questions to resolve

- Action mode: joint position, joint position + velocity, absolute EEF pose, or EEF pose delta.
- Command reference frame: world, robot base, or local tool frame; include frame and TCP identities.
- Delta semantics: translation frame, rotation composition order, and whether increments are relative to the measured or previous target pose.
- Pose encoding, quaternion ordering, units, and normalization conventions.
- Control period, interpolation/hold behavior, and sample timestamps.
- Gripper command representation and multi-arm endpoint ordering.
- Controller/IK configuration and version, including relevant limits and mappings needed to interpret or replay the commands.
- Joint names, ordering, and schema version for backward-compatible loading.

#### Acceptance criteria

- [ ] Document when EEF pose belongs in the primary action, an observation, or an auxiliary target field.
- [ ] Define schemas for the chosen joint and Cartesian representations, including frame, rotation, gripper, and timing semantics.
- [ ] Verify that the proposed fields are compatible with the supported LeRobot export and training/loading interfaces.
- [ ] Provide a small joint-controlled and Cartesian-controlled example showing observation/action alignment and the deployment controller boundary.
- [ ] Add focused encoding/loading tests if implemented, including multi-arm ordering and a frame or delta-semantics mismatch case.
- [ ] Preserve the existing joint-action dataset default and document any migration required for new representations.

### Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

yuecideng:
## Description

This PR defines an explicit joint/EEF action contract for EmbodiChain expert and LeRobot datasets while keeping joint position as the default representation.

It preserves raw requested EEF commands through the action manager, records measured FK EEF poses as auxiliary observations, records executed controller qpos separately, and adds the Franka RLinf-facing repeated pick-and-place task variants. The joint gripper term also exposes its true 8D policy action space, including normalized gripper bounds. The contract and temporal alignment are documented in the task README and episode metadata.

This PR intentionally does not include RLinf training/runtime fixes, joint dataset regeneration or SFT checkpoints, or DexSim dependency changes. Those belong to follow-up changes after this contract is available.

Fixes #601

## Type of change

- [ ] Documentation update

## Validation

- EmbodiChain focused tests: 55 passed.
- RLinf EmbodiChain config tests: 5 passed in the RLinf environment.
- Public API coverage: 2261/2261 exports documented.
- API documentation tests: 8 passed.
- Full-tree Black check, compileall, and `git diff --check` passed.

## Checklist

- [x] I reviewed affected documentation and agent context; no context guidance required a change.
- [x] I have added tests that prove the action contract and recorder behavior.


dustinCodes84600:
In EEF mode, raw history is captured before `ActionManager.process_action`, but `EefPoseTerm` applies `_scale` before IK. With any non-1 scale, dataset `action` and metadata claim an absolute arena-frame target while storing different values from the target actually sent to IK; same mismatch can happen for an out-of-range gripper value that execution clamps. Might be safer to record the processed `eef_pose` field alongside qpos, or explicitly define metadata as pre-scale policy input.

made a visual pass while reading through the diff: https://flyovers.dev/flyover/bb11f5c8-a043-4177-8fcd-35ae9851bc12

